PR #7, opened 2025-04-16
Add DynamicallyAccessedMember annotations for AOT

agocke:
Azure MCP uses some very simple reflection, which is compatible with AOT. However, AOT requires that the reflection info is propagated up so that the compiler can determine what info is used. The DynamicallyAccessedMembersAttribute does that by acting like a generic constraint, and pushing the info up the call stack to the substitution point.


jongio (approved):

conniey (approved):

conniey:
Apologies. I thought I could re-open the PR after closing it (and some magic would happen) because it will not let me merge, even with 2 approved reviewers.  @agocke would you be able to rebase your changes on top of this new main or open another PR?

agocke:
Yup happy to

PR #16, opened 2025-04-18
Add Dockerfile

jdubois:
Add a Dockerfile so the project can be built and run with Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,11 @@ RUN apt-get update && apt-get install -y \
 COPY --from=publish /app/publish .
 
 # Set environment variables with defaults that can be overridden
-ENV AZMCP_TRANSPORT=sse
+# the tranport can be set to "stdio" or "sse".
+# "stdio" is the default mode and uses standard input/output for communication.
+# "sse" is the server-sent events mode and uses HTTP for communication.
+ENV AZMCP_TRANSPORT=stdio
+# The port is used by the "sse" transport mode only, and defaults to 5008.
 ENV AZMCP_PORT=5008
 
 # Create entrypoint script
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+# Use the official .NET SDK image as the build image
+FROM mcr.microsoft.com/dotnet/sdk:9.0.102-bookworm-slim AS build
+WORKDIR /src
+
+# Copy csproj and restore dependencies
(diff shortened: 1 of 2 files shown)


dgageot:
Works just fine, as far as I can tell

conniey:
I tried the dockerfile and I don't think it's what we want the docker image to do.  We want it to start the MCP server.  Running `dotnet azmcp` by itself ends up showing a help message because there is a CLI mode for azmcp, too.

We would want the docker image to run `azmcp server start --transport sse --port {your port}`.  By default the port is 5008 if nothing is specified.

jdubois:
I've added the `server start --transport sse` part, but there's no goal to have this run in `stdio` mode by default? This is the only one supported by Claude Desktop, so I'm a bit surprised.
Concerning the port, I didn't specify it, as in Docker you can map the default 5008 port to whatever you want.

conniey:
> I've added the `server start --transport sse` part, but there's no goal to have this run in `stdio` mode by default? This is the only one supported by Claude Desktop, so I'm a bit surprised. Concerning the port, I didn't specify it, as in Docker you can map the default 5008 port to whatever you want.

We can have it run in stdio by default. :)

jdubois:
@conniey I added 2 env variables, `AZMCP_TRANSPORT` and `AZMCP_PORT` so those can be parametrized by the end user, is this OK for you?

conniey:
> @conniey I added 2 env variables, `AZMCP_TRANSPORT` and `AZMCP_PORT` so those can be parametrized by the end user, is this OK for you?

Sorry, there has been some miscommunication. STDIO for the docker image is fine. :)  The Dockerfile works great going back to the previous commit and changing: 
[`CMD ["server", "start", "--transport", "sse"]`](https://github.com/Azure/azure-mcp/pull/16/commits/0d6f570480cd5fe27545207192b7f16ad778696f#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R38)
to
`CMD ["server", "start" ]`

Thank you again for your contribution!

jdubois:
@conniey I modified the Dockerfile to have `stdio` by default, but kept the environment variables to easily set up the `sse` transport & port, and documented it. Is it fine like this?

conniey (approved):
Thank you!

conniey:
This could help with #64

PR #22, opened 2025-04-20
Async credentials retrieval

danielmarbach:
This small contribution makes the credential retrieval async to avoid the `Task.Run` and the `GetAwaiter() `call which seems unnecessary.


christothes (approved):
Thanks @danielmarbach ! This Task.Run code was left over from a previous factoring of this code that was designed to pre-cache some credentials in the background in a non-blocking way. It's not necessary in the current version, so thanks for this cleanup :)

jongio (approved):

PR #23, opened 2025-04-20
Switch WindowHandleProvider to library import and support MacOS

danielmarbach:
I tested the WindowProvider code on my MacOS (latest) and it works

### DIFF
--- a/tests/Services/Azure/Authentication/WindowHandleProvider.cs
+++ b/tests/Services/Azure/Authentication/WindowHandleProvider.cs
@@ -8,28 +8,10 @@ namespace AzureMcp.Tests.Services.Azure.Authentication;
 /// <summary>
 /// Provides window handle information for native authentication dialogs.
 /// </summary>
-public static class WindowHandleProvider
+public static partial class WindowHandleProvider
 {
     /// <summary>
-    /// Get the handle of the foreground window for Windows
-    /// </summary>
-    [DllImport("user32.dll")]
-    static extern IntPtr GetForegroundWindow();
-
-    /// <summary>
-    /// Get the handle of the console window for Linux
-    /// </summary>
-    [DllImport("libX11.so.6")]
-    private static extern IntPtr XOpenDisplay(string display);
-
-    [DllImport("libX11.so.6")]
-    private static extern IntPtr XRootWindow(IntPtr display, int screen);
-
-    [DllImport("libX11.so.6")]
-    private static extern IntPtr XDefaultRootWindow(IntPtr display);
-
-    /// <summary>
-    /// Get window handle on xplat
+    /// Get window handle across platforms
     /// </summary>
     public static IntPtr GetWindowHandle()
     {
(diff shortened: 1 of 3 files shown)


tests/Services/Azure/Authentication/WindowHandleProvider.cs
```diff
@@ -60,9 +43,74 @@ public static IntPtr GetWindowHandle()
             }
             return IntPtr.Zero;
         }
-        else
+
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
christothes, line 47: Thanks for the contribution, @danielmarbach 
The purpose of this window handle code is to ensure that the broker can appear above the z-order of the current application window. It is my understanding that MacOS didn't support the concept of inter-process window handles being used for z-order. Is there something I could be missing here?

As a side note, we don't actually support the broker on MacOS yet in any case. It should be coming soon though.

danielmarbach: I'm not super deep into this. When executing the tests on macos the interactive tests ran into the throwing of the platform exception part. After adding the platform specific part the interactive tests started working with the browser popping up and asking for the credentials. That's all I verified

christothes: Thanks, makes sense. You are seeing the fallback to the interactive browser flow, not the broker. 

I think perhaps the best change to make short term would be to simply Assert.Ignore in the test if `true == RuntimeInformation.IsOSPlatform(OSPlatform.OSX)`

danielmarbach: That would mean updating all the usages of this class do have the Assert.Ignore, remove the macos branch here and leave the library import. Did I understand your proposal correctly?

christothes: The only usage is the test, as far as I can see. I'm suggesting we revert the changes to WindowHandleProvider and assert.Ignore in the test that references this when running on MacOS. 

Once we support broker on MacOS, we'll need to figure out what the right thing to do here should be. As far as I understand, we would probably only supply IntPtr.Zero for MacOS, since the window handle won't be used by the broker APIs.

danielmarbach: Why should the source Gen library import usage be reverted? That seems a valuable change we would loose by reverting the class

danielmarbach: @christothes I pushed an update. Unfortunately, this project still targets xunit 2 and therefore doesn't allow adding skippable facts without taking a dependency to a community maintained project. Extending the existing fact attribute also felt a bit like scope creep...

danielmarbach: @christothes see https://github.com/Azure/azure-mcp/pull/31

christothes (approved):

jongio (approved):

PR #37, opened 2025-04-22
Add command `sql server database query`

stuartpa:
# Summary

- Adds an initial wave of SQL Database management and query commands under the Sql command group.
- Implements commands for executing queries, listing databases, and managing SQL resources in Azure.
- Ensures all commands use primary constructors, static members where possible, and System.Text.Json for serialization.

Sample usage

```
azmcp sql server list
azmcp sql server database list --server-name <server-name> --subscription <guid>
azmcp sql server database query --server-name <server-name> --subscription <guid> --database-name <db-name> --query "select * from sys.tables"
```

# Changes

- Introduces the SqlDatabaseQueryCommand for executing SQL statements against Azure SQL Databases.
- Adds argument and option definitions for SQL server, database, and query parameters.
- Implements command registration and argument binding for all SQL-related commands.
- Serializes command results using System.Text.Json with indented formatting.
- Follows repository conventions for command structure, file organization, and static member usage.

# Notes

- All new classes and interfaces are placed in separate files.
- Adheres to user-provided coding instructions and repository standards.
- No breaking changes introduced; new functionality is additive.

# Testing

- Added and updated unit tests for all new SQL commands:
  - SqlDatabaseQueryCommandTests covers query execution and result serialization.
  - SqlDatabaseListCommandTests validates listing of SQL databases.
  - SqlServerListCommandTests ensures correct server listing functionality.
  - All tests are located in the verify command behavior, argument parsing, and output formatting.
  - Manual validation of command execution and output.

This PR builds on #35, #36, it is tracked by Issue #33


docs/azmcp-commands.md
```diff
@@ -18,18 +19,24 @@ The following args are available for all commands:
 ## Available Commands
 
 ### Server Operations
+---
```
conniey, line 22: Are these needed?

eng/scripts/Build-Local.ps1
```diff
@@ -11,6 +11,9 @@ param(
     [switch] $VerifyNpx
 )
 
+# Add C:\Program Files\nodejs\npx to path
```
conniey, line 14: This script can be run cross platform and this ties it to Windows.

src/Commands/Sql/SqlDatabaseListCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.CommandLine;
+using System.CommandLine.Parsing;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace AzureMcp.Commands.Sql;
+
+public sealed class SqlDatabaseListCommand() : BaseCommand
+{
+    private static readonly Option<string> SubscriptionOption = new Option<string>(
```
conniey, line 15: These would be first declared in AzureMcp.Models.Argument.ArgumentDefinitions and the ToOption() extension method is used.

src/Commands/Sql/SqlDatabaseQueryCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+using Azure.ResourceManager;
+using Azure.ResourceManager.Sql;
+using Azure.ResourceManager.Resources;
+using Microsoft.Data.SqlClient;
+
+namespace AzureMcp.Commands.Sql;
+
+public sealed class SqlDatabaseQueryCommand(ISqlDatabaseQueryService queryService) : BaseCommand
+{
+    public static Option<string> SubscriptionOption { get; } = new Option<string>(
```
conniey, line 16: Similarly with the comment about ArgumentDefinitions above.

src/Commands/Sql/SqlDatabaseQueryService.cs
```diff
@@ -0,0 +1,50 @@
...
+        if (string.IsNullOrEmpty(fqdn))
+            return new { Error = "SQL Server not found." };
+
+        var builder = new SqlConnectionStringBuilder
+        {
+            DataSource = fqdn,
+            InitialCatalog = databaseName,
+            Authentication = SqlAuthenticationMethod.ActiveDirectoryDefault
+        };
+        var results = new List<Dictionary<string, object?>>();
```
conniey, line 36: I wonder if this would benefit from progress reporting (https://modelcontextprotocol.io/docs/concepts/tools#best-practices) because it seems like it could be a long running operation.

src/Commands/Sql/SqlDatabaseQueryService.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.Data.SqlClient;
+using System.Collections.Generic;
+
+namespace AzureMcp.Commands.Sql;
+
+public sealed class SqlDatabaseQueryService() : ISqlDatabaseQueryService
+{
+    public async Task<object> ExecuteQueryAsync(string subscription, string serverName, string databaseName, string query)
+    {
+        var credential = new DefaultAzureCredential();
```
conniey, line 15: This should extend from BaseAzureService, we want to use GetCredential() method in this class to fetch the credential.  (Getting credentials is an expensive operation.)

Similarly with the other services.

docs/azmcp-commands.md
stuartpa: They get added by VSCode automatically.  I'll remove them now to cleanup the PR

eng/scripts/Build-Local.ps1
stuartpa: My mistake, this was a temporary line I add to get going.  Removed.

src/Commands/Sql/SqlDatabaseListCommand.cs
stuartpa: Fixed

src/Commands/Sql/SqlDatabaseQueryService.cs
stuartpa: Fixed

src/Commands/Sql/SqlDatabaseQueryService.cs
stuartpa: Great feedback, fixed

joshfree:
Let's trim this down to the 2-6 Pri-0 scenarios, ideally READ-ONLY, for the Build-2 timeframe (which is the next ~2.5 weeks).

Let's defer the other lower priority commands until AFTER the dynamic tool loading changes from @conniey are in the product.  Otherwise this will push us over the 40 tool limit which will cause problems for our end users until dynamic tool loading is in.

AV25242:
Bringing over my questions from chat to here as advised

1. I don't see an implementation of [Service Interface or Service class](https://github.com/Azure/azure-mcp/blob/main/src/Docs/new-command.md#step-2-define-service-interface-method) like described in the documentation. Should this be a part of the implementation for SQL tools to be picked up when the client makes a list tool call?
2. How will or should the cli azmcp be used. Is it to simply test the behavior of the tool or its used internally by Azure MCP server to load and list the tools?
3. In the PR I see Commands/SQL has both the command and arguments file, probably the arguments live in its own Arguments folder?

docs/azmcp-commands.md
```diff
@@ -48,6 +48,21 @@ azmcp cosmos database container item query --subscription <subscription> \
                        [--query "SELECT * FROM c"]
 ```
 
+### Sql Operations
+```bash
+# List SQL Servers in a subscription
+azmcp sql server list --subscription <subscription>
+
+# List SQL Databases in a SQL Server
+azmcp sql server database list --subscription <subscription> --server-name <server-name>
```
xiangyan99, line 57: why not just use azmcp sql database list which matches cli?

stuartpa: I followed the CosmosDB pattern for azcmp, which puts container under database (the equivalent for Sql is database under server).  It's the same in the Az CLI for CosmosDB, its:

az cosmosdb sql database
az cosmosdb sql container

But their azcmp command  is

azcmp cosmodb database container

The equivalent for SQL in the az cli is (a db is contained within a server)

az sql server
az sql db

So to be the same as azcmp cosmosdb, for sql we end up with

azcmp sql server database

(I expanded out the word "database" from "db" to again be the same as CosmosDB in azcmp)

Is it correct for sql to follow the CosmosDB pattern in azcmp, or to follow the pattern in Az CLI (which cosmosdb doesn't follow)

xiangyan99: It seems to me that CosmosDB follows the same pattern as what's used in the CLI,  https://learn.microsoft.com/en-us/cli/azure/cosmosdb?view=azure-cli-latest

whereas SQL uses a different pattern in your CLI command. :)

stuartpa: Can you give an example.  I have given you an example of how SQL follows the same pattern as Cosmos, again:

Cosmos:

azmcp cosmos database container item query

Sql:

azmcp sql server database query

xiangyan99: This is the container command in the Cosmos CLI:
image

And here’s your command:
image

To stay consistent with the existing pattern, I don't think we need "server" in your command.
We could simply use:
"azmcp sql database query".

Are you planning to introduce something like:
azmcp sql xxx database query?

If not, I believe "azmcp sql database query" is sufficient.

xiangyan99: Thanks for the information.

In CosmosDB, operations can be performed directly on databases:
image

Do you currently have, or plan to add, operations on servers as well?
(I couldn’t find any server-related operations in the CLI.)

jongio:
@stuartpa - We can make this more clear, but you should be able to feed the new-command.md file into copilot and have it generate all the things for you    "create a new command xyz using #new-command.md as a reference"

xiangyan99 (approved):

xiangyan99:
Please rebase main.

src/Commands/Sql/SqlDatabaseListArguments.cs
```diff
@@ -0,0 +1,7 @@
+namespace AzureMcp.Commands.Sql;
+
+public sealed class SqlDatabaseListArguments(string? Subscription, string? ServerName)
+{
+    public string? Subscription { get; } = Subscription;
+    public string? ServerName { get; } = ServerName;
+}
```
xiangyan99, line 7: Please put the arguments into src/Arguments/{Service}/{SubService}/{Resource}/{Resource}{Operation}Arguments.cs

src/Commands/Sql/ISqlServerLister.cs
```diff
@@ -0,0 +1,6 @@
+namespace AzureMcp.Commands.Sql;
```
xiangyan99, line 1: Please put the interface into src/Services/Interfaces/I{Service}Service.cs

xiangyan99: You can get more details in https://github.com/Azure/azure-mcp/blob/main/src/Docs/new-command.md

src/Commands/Sql/SqlServerListCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (parts[i].Equals("resourceGroups", System.StringComparison.OrdinalIgnoreCase) && i + 1 < parts.Length)
+                return parts[i + 1];
+        }
+        return string.Empty;
+    }
+
+    public static async Task<IReadOnlyList<object>> ListSqlServersAsync()
+    {
+        var credential = new DefaultAzureCredential();
```
xiangyan99, line 43: We should use GetCredential from BaseAzureService

src/Commands/Sql/SqlServerLister.cs
```diff
@@ -0,0 +1,43 @@
+using Azure.Identity;
+using Azure.ResourceManager;
+using Azure.ResourceManager.Sql;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace AzureMcp.Commands.Sql;
+
+public sealed class SqlServerLister() : ISqlServerLister
```
xiangyan99, line 9: It seems to me we should have a SqlDatabaseService and this should be part of the service.

xiangyan99:
I don't see you register your service in ConfigureServices?

xiangyan99 (changes requested):
Please rebase the main and re-organize the files.

src/Commands/Sql/SqlDatabaseQueryService.cs
```diff
@@ -0,0 +1,82 @@
...
+using Azure.ResourceManager.Resources;
+using Azure.ResourceManager.Sql;
+using AzureMcp.Services.Azure;
+using Microsoft.Data.SqlClient;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace AzureMcp.Commands.Sql;
+
+public sealed class SqlDatabaseQueryService() : BaseAzureService, ISqlDatabaseQueryService
```
xiangyan99, line 12: Looks like we need ISubscriptionService here

stuartpa:
> @stuartpa - We can make this more clear, but you should be able to feed the new-command.md file into copilot and have it generate all the things for you "create a new command xyz using #new-command.md as a reference"

I did do this, and this PR is what it produced!

stuartpa:
Closing for now, we look after //BUILD

dennis-yemelyanov:
@stuartpa , is there any plan to complete this PR? It would be very helpful to have sql query supported for various automation scenarios

stuartpa:
@dennis-yemelyanov , @AV25242 (PM in this space) is on point from a prioritization point of view for this work.

dennis-yemelyanov:
@stuartpa , yes, they seem to support queries, but there is one problem, that the server/database name need to be passed when starting the MCP server and can't be customized later. We need to query different databases and can't really start a separate server for each of them.

If they can be made more flexible to accept server/database name as function parameters (like this repository does), we would be happy to try them out

AV25242:
> Thank you. @AV25242 , any thoughts on this? We are trying to decide whether we can use something that will be available soon or need to build our own MCP for SQL queries

Hi @dennis-yemelyanov the Azure MCP for Azure SQL will only have the management plane operations. The data plane operations will be handled by mssql mcp server like @stuartpa mentioned and this is because we want a single unified MCP Server for SQL Server's data ops.

Coming back to your request can you create a feature request using [GitHub Issues](https://github.com/Azure-Samples/SQL-AI-samples/issues) explaining exactly what you are looking for and the use case.

PR #40, opened 2025-04-23
Add Compute VM list functionality

Reedboot:
This PR adds Compute VM list functionality and can be expanded upon by others.


Reedboot:
@microsoft-github-policy-service agree

src/Commands/CommandFactory.cs
```diff
@@ -1,12 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using AzureMcp.Commands.Compute.Vm;
```
jongio, line 4: Make sure you run `dotnet format src/AzureMcp.sln`

src/Commands/Compute/Vm/VmListCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+using ModelContextProtocol.Server;
+using System.CommandLine.Parsing;
+
+namespace AzureMcp.Commands.Compute.Vm;
+
+public sealed class VmListCommand : SubscriptionCommand<VmListArguments>
+{
+    private readonly ILogger<VmListCommand> _logger;
+
+    public VmListCommand(ILogger<VmListCommand> logger) : base()
```
jongio, line 17: Use primary ctor

src/AzureMcp.csproj
```diff
@@ -70,5 +71,9 @@
     <PackageReference Include="Azure.Monitor.OpenTelemetry.AspNetCore" Version="1.2.0" />
     <PackageReference Include="Microsoft.Extensions.Azure" Version="1.11.0" />
   </ItemGroup>
+
+  <ItemGroup>
```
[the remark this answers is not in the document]
Reedboot: The item group tag or the reference to the folder?

jongio:
@Reedboot - we are evaluating whether or not to add more mgmt plane ops like this one b/c we recently integrated the azure cli tool, which can cover a lot of these commands already.  for example

If i issue "list vms in sub [subid]" it will use the azmcp-extension-az tool and successfully get them.

do you have any ideas for data plane related tools for vms or other services

Reedboot:
It does sound redundant then with the Cli.

Happy to go whatever way you feel is best with this.

My concern however with allowing Cli is that could allow delete and modify of existing azure assets where a method approach is more restricted. 

Is there a separate chat / group that I can join to discuss further?

jongio:
Thanks, I'll close this for now and we can revisit if we have data plane ops to add.  Are you internal? If so, you can ping me on Teams for internal chat.  I'll ask to enable discussions on this repo

Reedboot:
@jongio not internal, just a member of the public trying to help.

PR #44, opened 2025-04-23
feat: more robust az path resolution

woutervanranst:
When Az CLI is installed via `pip` on Windows, it will be in `Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "Programs", "Python", "Python39", "Scripts"),` as both `az` and `az.bat`. The proposed changes look these up in a more robust and non breaking manner by 1/ when `isWindows` also checking for `az.bat` and regardless of OS also checking for `az`

### DIFF
--- a/src/Commands/Extension/AzCommand.cs
+++ b/src/Commands/Extension/AzCommand.cs
@@ -97,15 +97,20 @@ Your job is to answer questions about an Azure environment by executing Azure CL
                     _cachedAzPath = cmdPath;
                     return cmdPath;
                 }
-            }
-            else
-            {
-                var fullPath = Path.Combine(dir, "az");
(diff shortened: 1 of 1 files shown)


xiangyan99:
Thanks for your contribution and for bringing this to our attention.

I've created a PR (https://github.com/Azure/azure-mcp/pull/63) to refactor our implementation and address the issue you raised. When you have a chance, could you please take a look and let me know if it solves your problem?

Your feedback would be greatly appreciated!

jongio:
@xiangyan99 can we close this now?

xiangyan99:
Thank for your contribution. We made the refactoring work in https://github.com/Azure/azure-mcp/pull/63. Closing the issue.

PR #51, opened 2025-04-24
Add DynamicallyAccessedMember annotations for AOT

agocke:
Azure MCP uses some very simple reflection, which is compatible with AOT. However, AOT requires that the reflection info is propagated up so that the compiler can determine what info is used. The DynamicallyAccessedMembersAttribute does that by acting like a generic constraint, and pushing the info up the call stack to the substitution point.


joshfree (approved):
Thanks @agocke

PR #52, opened 2025-04-25
Use source-generated JSON serialization

agocke:
This rewrites all the reflection-based serialization to source-generation for AOT compat. In addition, we can now enable the AOT analyzers.

This doesn't solve warnings in referenced libraries, but it should make this server fully AOT compatible.


src/Commands/JsonSrcGenCtx.cs
```diff
@@ -0,0 +1,53 @@
...
+[JsonSerializable(typeof(KeyValueDeleteCommand.KeyValueDeleteCommandResult))]
+[JsonSerializable(typeof(ContainerDetailsCommand.ContainerDetailsCommandResult))]
+[JsonSerializable(typeof(ItemQueryCommand.ItemQueryCommandResult))]
+[JsonSerializable(typeof(JsonElement))]
+[JsonSerializable(typeof(List<string>))]
+[JsonSerializable(typeof(List<JsonNode>))]
+[JsonSerializable(typeof(CommandResponse))]
+[JsonSerializable(typeof(AzureMcp.Models.ETag), TypeInfoPropertyName = "McpETag")]
+[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+internal partial class JsonSrcGenCtx : JsonSerializerContext
```
conniey, line 50: We generally avoid using abbreviations for names

src/Models/Command/CommandResponse.cs
```diff
@@ -20,8 +22,45 @@ public class CommandResponse
...
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public object? Results { get; set; }
+    public ResponseResult? Results { get; set; }
 
     [JsonPropertyName("duration")]
     public long Duration { get; set; }
+}
+
+[JsonConverter(typeof(ResultConverter))]
+public sealed class ResponseResult
```
conniey, line 32: Consider moving this into a separate file?

joshfree:
Related https://github.com/Azure/azure-mcp-pr/issues/27

src/Models/Command/CommandResponse.cs
hallipr: Should we change the invocation style to hide the use of the local serialization class?

```cs
context.Response.SetResult(result, JsonSourceGenerationContext.Default.KeyValueDeleteCommandResult);
```

hallipr:
```cs
if (items?.Count > 0)
{
    context.Response.SetResult(new ItemQueryCommandResult(items), JsonSourceGenerationContext.Default.ItemQueryCommandResult);
} else {
    context.Response.SetResult(null);
}
```

src/Commands/Tools/ToolsListCommand.cs
```diff
@@ -12,16 +12,21 @@
 namespace AzureMcp.Commands.Tools;
 
 [HiddenCommand]
-public sealed class ToolsListCommand(ILogger<ToolsListCommand> logger) : BaseCommand()
+public sealed class ToolsListCommand : BaseCommand
```
jongio, line 15: Let's use primary ctors please

src/Commands/JsonSrcGenCtx.cs
```diff
@@ -0,0 +1,53 @@
...
+[JsonSerializable(typeof(KeyValueDeleteCommand.KeyValueDeleteCommandResult))]
+[JsonSerializable(typeof(ContainerDetailsCommand.ContainerDetailsCommandResult))]
+[JsonSerializable(typeof(ItemQueryCommand.ItemQueryCommandResult))]
+[JsonSerializable(typeof(JsonElement))]
+[JsonSerializable(typeof(List<string>))]
+[JsonSerializable(typeof(List<JsonNode>))]
+[JsonSerializable(typeof(CommandResponse))]
+[JsonSerializable(typeof(AzureMcp.Models.ETag), TypeInfoPropertyName = "McpETag")]
+[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+internal partial class JsonSourceGenerationContext : JsonSerializerContext
```
danielmarbach, line 50: Given some of those types come from specific "feature" areas of the MCP server, would it make sense to split the partial definition up across the folders?

src/Models/Command/CommandResponse.cs
agocke: I could do that, but I usually keep types that are serialized as simple as possible. If I add a SetResult API I assume I would need to make the Response private, so it wouldn't be accessible. But then it wouldn't be serialized automatically, right?

src/Commands/Tools/ToolsListCommand.cs
agocke: Ah, can't remember why I cahnged that. I will revert.

src/Commands/JsonSrcGenCtx.cs
agocke: I don't see a problem with that. I can split every place where the type names collide.

jongio:
@agocke - Can you rebase?

conniey (approved):
Thank you!

hallipr (approved):

PR #61, opened 2025-04-28
Allow Azmcp to grab auth session from VS Code

JasonYeMSFT:
Major changes:

1. Set UseDefaultBrokerAccount to false when using the "Broker Credential only" option or an authentication session is present.
2. Modified the code so the InteractiveBrowserCredential can specify a tenant
3. Added code to grab the parent window handle so azmcp can open a system dialog for interactive browser authentication
4. Added an environment variable which indicates only using the Broker Credential
5. Added an environment variable to allow overriding the client id used in the InteractiveBrowserCredential
6. Added an environment variable to pass in an authentication record
7. Added TokenCacheOptions


joshfree:
Required for `2025-05`

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,99 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+    static extern IntPtr GetAncestor(IntPtr hwnd, GetAncestorFlags flags);
 
-    private static InteractiveBrowserCredential CreateBrowserCredential() =>
-        new(new InteractiveBrowserCredentialBrokerOptions(IntPtr.Zero)
+    [DllImport("kernel32.dll")]
+    static extern IntPtr GetConsoleWindow();
+
+    public static IntPtr GetConsoleOrTerminalWindow()
+    {
+        if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
JasonYeMSFT, line 82: I talked to @hallipr who worked on the build and publishing process and we decided to use runtime detection for now. Previously I was using a "WINDOWS" symbol to do per-platfrom compliation. It could work but it is much more complicated than this because we will need additional changes to make tests run in platform specific modes and allow local dev environments to automatically use the correct mode.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,99 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+        if (!string.IsNullOrEmpty(authRecordJson))
+        {
+            byte[] bytes = Encoding.UTF8.GetBytes(authRecordJson);
+            using MemoryStream authRecordStream = new MemoryStream(bytes);
+            authRecord = AuthenticationRecord.Deserialize(authRecordStream);
+        }
 
-    public override ValueTask<AccessToken> GetTokenAsync(TokenRequestContext requestContext, CancellationToken cancellationToken) =>
-        _chainedCredential.GetTokenAsync(requestContext, cancellationToken);
+        string? useOnlyBrokerCredential = Environment.GetEnvironmentVariable("AZURE_MCP_USE_ONLY_BROKER_CREDENTIAL");
```
jongio, line 46:
```suggestion
        string? useOnlyBrokerCredential = Environment.GetEnvironmentVariable("AZURE_MCP_ONLY_USE_BROKER_CREDENTIAL");
```

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,99 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+        if (useOnlyBrokerCredential == "true")
```
jongio, line 47: Have better truthy check, true, True, T, 1

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,99 @@ namespace AzureMcp.Services.Azure.Authentication;
...
-        new(CreateDefaultCredential(tenantId), CreateBrowserCredential());
+    /// <summary>
+    /// Retrieves the handle to the ancestor of the specified window.
+    /// </summary>
+    /// <param name="hwnd">A handle to the window whose ancestor is to be retrieved.
+    /// If this parameter is the desktop window, the function returns NULL. </param>
+    /// <param name="flags">The ancestor to be retrieved.</param>
+    /// <returns>The return value is the handle to the ancestor window.</returns>
+    [DllImport("user32.dll", ExactSpelling = true)]
+    static extern IntPtr GetAncestor(IntPtr hwnd, GetAncestorFlags flags);
```
jongio, line 75: You should be using the WindowHandler that is in the tests folder.  You can bring it into product /src folder if need be.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,99 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+            return IntPtr.Zero;
+        }
+    }
+
+    private static string TokenCacheName = "azure-mcp-msal.cache";
+
+    private static InteractiveBrowserCredential CreateBrowserCredential(string? tenantId, AuthenticationRecord? authRecord)
+    {
+        string? clientId = Environment.GetEnvironmentVariable("AZURE_MCP_AUTH_CLIENT_ID");
+        string? useOnlyBrokerCredential = Environment.GetEnvironmentVariable("AZURE_MCP_USE_ONLY_BROKER_CREDENTIAL");
```
jongio, line 100: Don't repeat strings

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,99 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+
+        IntPtr handle = GetConsoleOrTerminalWindow();
+
+        return new(new InteractiveBrowserCredentialBrokerOptions(handle)
         {
-            UseDefaultBrokerAccount = true
+            UseDefaultBrokerAccount = useOnlyBrokerCredential != "true" && authRecord is null,
```
jongio, line 106: refactor truthy check

src/Services/Azure/Authentication/CustomChainedCredential.cs
JasonYeMSFT: Not needed anymore since there are existing code I can reuse to get window handle.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +19,72 @@ namespace AzureMcp.Services.Azure.Authentication;
 /// </remarks>
 public class CustomChainedCredential(string? tenantId = null) : TokenCredential
 {
-    private readonly ChainedTokenCredential _chainedCredential = CreateChainedCredential(tenantId);
+    public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
+    {
+        ChainedTokenCredential chainedCredential = CreateChainedCredential(tenantId);
```
christothes, line 24: Do we need to create the credential instance each time GetToken is called, or is it possible to cache this?

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +19,72 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+        string? onlyUseBrokerCredential = Environment.GetEnvironmentVariable(OnlyUseBrokerCredentialEnvVarName);
+        return onlyUseBrokerCredential == "true"
+            || onlyUseBrokerCredential == "True"
+            || onlyUseBrokerCredential == "T"
+            || onlyUseBrokerCredential == "1";
+    }
 
-    public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken) =>
-        _chainedCredential.GetToken(requestContext, cancellationToken);
+    private static ChainedTokenCredential CreateChainedCredential(string? tenantId)
```
christothes, line 47: nit: We could have this return `TokenCredential` and name it `GetCredential` so that the implementation is agnostic.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +19,72 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+    private const string ClientIdEnvVarName = "AZURE_MCP_CLIENT_ID";
+
+    private static bool ShouldUseOnlyBrokerCredential()
+    {
+        string? onlyUseBrokerCredential = Environment.GetEnvironmentVariable(OnlyUseBrokerCredentialEnvVarName);
+        return onlyUseBrokerCredential == "true"
+            || onlyUseBrokerCredential == "True"
+            || onlyUseBrokerCredential == "T"
+            || onlyUseBrokerCredential == "1";
+    }
```
christothes, line 45: nit: slightly more compact
```suggestion
    private static bool ShouldUseOnlyBrokerCredential() =>
        Environment.GetEnvironmentVariable(OnlyUseBrokerCredentialEnvVarName) switch
        {
            "true" => true,
            "True" => true,
            "T" => true,
            "1" => true,
            _ => false
        };
```

christothes (approved):

src/Services/Azure/Authentication/CustomChainedCredential.cs
JasonYeMSFT: Do we need to worry about concurrent access to the _chainedCredential instance? Or is azmcp implemented to be essentially single threaded for now?

src/Services/Azure/Authentication/CustomChainedCredential.cs
jongio: This should be in a helper to be reused (if one doesn't already exist in .NET)

src/Services/Azure/Authentication/WindowHandleProvider.cs
jongio, line 1: You are also removing the file from the /tests folder right?

JasonYeMSFT: Yes, I removed the corresponding file from the /tests folder. I also just learned that GitHub UI is different from ADO. It shows this as a file move rather than a delete + an addition.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: The chained token credential is using impl from Identity SDK which is thread safe by design.

src/Services/Azure/Authentication/CustomChainedCredential.cs
JasonYeMSFT: I refactored it to a helper class method `GetEnvironmentVariableAsBool` that can be reused in other classes.

src/Services/Azure/Authentication/CustomChainedCredential.cs
JasonYeMSFT: I updated the code in this PR to cache the credential object. Let's add synchronization in a follow up PR.

JasonYeMSFT:
@jongio @christothes @g2vinay Could you please review again?

christothes (approved):

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,70 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+    {
+        string? clientId = Environment.GetEnvironmentVariable(ClientIdEnvVarName);
+
+        IntPtr handle = WindowHandleProvider.GetWindowHandle();
+
+        return new(new InteractiveBrowserCredentialBrokerOptions(handle)
         {
-            UseDefaultBrokerAccount = true
+            UseDefaultBrokerAccount = !ShouldUseOnlyBrokerCredential() && authRecord is null,
+            ClientId = clientId,
```
g2vinay, line 78: Will this set a null Client ID here if there is nothing in the env ?
There's a default client ID set in Broker Options, will a null here override that ?
@christothes

christothes: Yes, null will override that here. So we should not set it to null if we don't intend that to flow through.

JasonYeMSFT: I pushed a commit to address this issue.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,75 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+    private static InteractiveBrowserCredential CreateBrowserCredential(string? tenantId, AuthenticationRecord? authRecord)
+    {
+        string? clientId = Environment.GetEnvironmentVariable(ClientIdEnvVarName);
+
+        IntPtr handle = WindowHandleProvider.GetWindowHandle();
+
+        InteractiveBrowserCredentialBrokerOptions brokerOptions = new(handle)
+        {
+            UseDefaultBrokerAccount = !ShouldUseOnlyBrokerCredential() && authRecord is null,
+            TenantId = tenantId,
```
g2vinay, line 78: We can update to
TenantId = string.IsNullOrEmpty(tenantId) ? null : tenantId,

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,75 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+            AuthenticationRecord = authRecord,
+            TokenCachePersistenceOptions = new TokenCachePersistenceOptions()
+            {
+                Name = TokenCacheName,
+            }
+        };
+        if (clientId is not null)
```
jongio, line 85: I believe a blank line is required before the if

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -18,22 +20,75 @@ namespace AzureMcp.Services.Azure.Authentication;
...
+        {
+            brokerOptions.ClientId = clientId;
+        }
+
+        return new(brokerOptions);
+    }
 
     private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
     {
```
jongio, line 94: Use the new truthy helper

src/AzureMcp.csproj
```diff
@@ -18,6 +18,7 @@
     <PublishTrimmed>false</PublishTrimmed>
     <IncludeNativeLibrariesForSelfExtract>true</IncludeNativeLibrariesForSelfExtract>
     <DebugType>embedded</DebugType>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
jongio, line 21: Why is this required?

JasonYeMSFT: This is required by the code to get the WindowHandle.

src/Services/Azure/Authentication/CustomChainedCredential.cs
JasonYeMSFT: If you are referring to the clientId, its logic is different from the logic of the truth helper.

jongio: image

g2vinay (approved):

PR #63, opened 2025-04-28
refactor finding cli path

xiangyan99:


src/Commands/Extension/AzCommand.cs
```diff
@@ -72,43 +64,28 @@ protected override AzArguments BindArguments(ParseResult parseResult)
 
     private static string? FindAzCliPath()
     {
+        string executableName = "az";
```
jongio, line 67: Is the correct on all plats, win, linux, mac.

src/Commands/Extension/AzCommand.cs
```diff
@@ -72,43 +64,28 @@ protected override AzArguments BindArguments(ParseResult parseResult)
...
 
-        var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
-        var searchPaths = new List<string>(AzureCliPaths);
-
-        // Add PATH environment directories
-        if (Environment.GetEnvironmentVariable("PATH")?.Split(Path.PathSeparator) is { } pathDirs)
-        {
-            searchPaths.AddRange(pathDirs);
-        }
+        var pathEnv = Environment.GetEnvironmentVariable("PATH");
```
jongio, line 75: Is there a reliable xplat 'where' command?

xiangyan99: I don't think there is a xplat "where" command. And it seems to me this is the c# solution for xplat 'where'. :)

woutervanranst: You could use `which` on linux but it the proposed refactor is cleaner / more xplat

src/Commands/Extension/AzCommand.cs
```diff
@@ -72,43 +64,28 @@ protected override AzArguments BindArguments(ParseResult parseResult)
...
-                var cmdPath = Path.Combine(dir, "az.cmd");
-                if (File.Exists(cmdPath))
-                {
-                    _cachedAzPath = cmdPath;
-                    return cmdPath;
-                }
-            }
-            else
+            string fullPath = Path.Combine(path.Trim(), executableName);
+            if (File.Exists(fullPath))
```
woutervanranst, line 83: There is an edge case where the running process would not have access to the path and this would throw an `UnauthorizedAccessException`

xiangyan99: There are valid scenarios where we would run an az command.

One case is using the CLI credential; as you mentioned, if that fails, it should raise an UnauthorizedAccessException.

Another scenario is running the az CLI command as an extension.
In this case, raising an UnauthorizedAccessException would not be appropriate.

Given that, I believe it should be up to the caller to decide which exception to throw.

src/Commands/Extension/AzCommand.cs
xiangyan99: Yes, this is xplat and also part of our implementation of the identity library.

joshfree:
tagging as `Do Not Merge` while this is Draft / WIP

conniey (approved):
LGTM

src/Commands/Extension/AzCommand.cs
```diff
@@ -72,43 +64,44 @@ protected override AzArguments BindArguments(ParseResult parseResult)
...
-                }
-            }
-            else
-            {
-                var fullPath = Path.Combine(dir, "az");
-                if (File.Exists(fullPath))
-                {
-                    _cachedAzPath = fullPath;
-                    return fullPath;
+                    string exePath = fullPath + ".cmd";
```
joshfree, line 88: use System.IO.Path.ChangeExtension() here instead of string concatenation please.

https://learn.microsoft.com/en-us/dotnet/api/system.io.path.changeextension?view=net-9.0

joshfree (changes requested):
It's safest to use the System.IO.Path APIs rather than direct string manipulation.  This will handle things better when there are UNC paths or NTFS reparse points or someone trying to do something bad with environment variables

joshfree (approved):

PR #73, opened 2025-04-29
realtime logging

roryp:
Suggested modifications of the implementation to support real-time streaming of logs.
Add a new method to the IExternalProcessService interface that supports callbacks for real-time output
Implement this method in the ExternalProcessService class
Update the AzdCommand to use this new method for commands like "azd up" that benefit from real-time logging


roryp:
assignment should be Rohit Ganguly

src/Models/Command/CommandContext.cs
```diff
@@ -45,4 +53,13 @@ public T GetService<T>() where T : class
...
         return _serviceProvider.GetRequiredService<T>();
     }
+    
+    /// <summary>
+    /// Sends real-time results to the client during long-running operations
+    /// </summary>
+    /// <param name="response">The response containing real-time results to send</param>
+    public void SendRealtimeResults(CommandResponse response)
+    {
+        _mcpServer?.SendRealtimeResults(response);
```
conniey, line 63: I feel like I am missing something because I don't see SendRealtimeResults defined in IMcpServer

src/Commands/Extension/AzdCommand.cs
```diff
@@ -89,7 +93,60 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
...
+                // Create a StringBuilder to collect the streamed output for the final result
+                var streamedOutput = new List<string>();
+                
+                // Define handlers for real-time output
+                void StdOutHandler(string line)
+                {
+                    streamedOutput.Add(line);
+                    
+                    // Send the output line as it arrives
+                    var streamingResponse = new CommandResponse
```
conniey, line 112: I may be mistaken but the MCP protocol doesn't currently contain a concept of partial results or streaming.  When you return the first result, the MCP client will think that is the response from calling the tool.

conniey:
Please correct me if I am incorrect, the PR is trying to fix a gap in the observability experience when running commands that may be long-lived.  The MCP protocol doesn't have a concept of partial success or streaming data at the moment. So, the first response published would be considered the answer to the azd/az command.

Have you considered using some of the observability tools/logging we have in the project to fill this gap?  

* https://github.com/Azure/azure-mcp/blob/main/TROUBLESHOOTING.md#observability-with-opentelemetry
* https://github.com/Azure/azure-mcp/blob/main/TROUBLESHOOTING.md#logging

src/Commands/Extension/AzdCommand.cs
roryp: It does https://mcp-framework.com/docs/Transports/http-stream-transport

roryp:
> To make sure we are on the same page, the PR is trying to fix a gap in the observability experience when running commands that may be long-lived.
> 
> The MCP protocol doesn't have a concept of partial success or streaming data at the moment. So, the first response published would be considered the answer to the azd/az command. Have you considered using some of the observability tools/logging we have in the project to fill this gap?
> 
> * https://github.com/Azure/azure-mcp/blob/main/TROUBLESHOOTING.md#observability-with-opentelemetry
> * https://github.com/Azure/azure-mcp/blob/main/TROUBLESHOOTING.md#logging

Streaming is supported https://mcp-framework.com/docs/Transports/http-stream-transport

src/Models/Command/CommandContext.cs
roryp: ive pushed the files i missed. the mcp server does support streaming please look at https://mcp-framework.com/docs/Transports/http-stream-transport/

joshfree:
Thanks @roryp for the contribution!  @lmolkova could you also take a quick peek?

conniey:
Yes, the transport is supported. I am referring to the [message format](https://modelcontextprotocol.io/docs/concepts/transports#message-format, JSON-RPC 2.0], they are communicating over that transport.

If a client sends a request, the server should only send one JSON response back with the corresponding id for the request.  That response can have an array of content, but it all lives inside one JSON object that is serialized. https://www.jsonrpc.org/specification#batch

In the case of a long operation like `azd up`, the MCP service would publish this JSON response after all the messages have been collected, and `azd up` has completed.  In the meantime, you don't see anything from the service.

That's one of the reasons why we added OpenTelemetry and logging (dotnet-trace/ETW), so you can view in real-time if the service is stuck or not.

jongio:
@roryp - Please detail what this PR is solving, what issues you have so we can plan and design.  Asking @stephentoub to also review for any C# SDK improvements needed.

jongio:
Marking as Do Not Merge until Toub has reviewed

src/Commands/Extension/AzdCommand.cs
stephentoub: > It does

That isn't about streaming the results of tools; see @halter73's comment here:
https://github.com/modelcontextprotocol/csharp-sdk/issues/358#issuecomment-2832612965

src/Services/Interfaces/MCP/AzureMcpServerExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <param name="response">The response containing real-time results to send</param>
+    public void SendRealtimeResults(CommandResponse response)
+    {
+        // Implementation of sending real-time results
+        // This depends on how your MCP server handles streaming results
+        
+        // For now, we'll just log that this method was called
+        Console.WriteLine($"Sending real-time results: {response.Message}");
+        
+        // TODO: Implement actual streaming functionality based on your specific requirements
```
stephentoub, line 25: I'm struggling to understand the purpose of this PR. What is the intention with regards to how data is propagated to the MCP client? Are you looking to use notifications, or something else?

roryp:
closes as per @rohit asking for a delay as the developer is one leave

SandraAhlgrimm:
Please reopen as a start to add azd as requested here #89 @joshfree

PR #79, opened 2025-04-30
Remove arguments from response

jongio:
Enhance command and argument handling by introducing a StringConverter for improved string processing, updating argument definitions for better clarity, and modifying property names to reflect their purpose more accurately.


conniey (approved):
🥂 lgtm

src/Commands/Tools/ToolsListCommand.cs
```diff
@@ -49,14 +48,15 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
 
     private static CommandInfo CreateCommand(string hyphenatedName, IBaseCommand command)
```
joshfree, line 49: We have a separate bug report that states hyphenated command names are invalid and we should be snake_casing. Do you want to fix that at the same time?

joshfree (approved):

PR #90, opened 2025-05-02
Live tests for Azure AI Search commands

pablocastro:
This PR adds live tests for Azure AI Search-related commands to validate things work against real services.

### DIFF
--- a/src/Services/Azure/Search/SearchService.cs
+++ b/src/Services/Azure/Search/SearchService.cs
@@ -18,7 +18,9 @@ public sealed class SearchService(ISubscriptionService subscriptionService, ICac
     private readonly ISubscriptionService _subscriptionService = subscriptionService ?? throw new ArgumentNullException(nameof(subscriptionService));
     private readonly ICacheService _cacheService = cacheService ?? throw new ArgumentNullException(nameof(cacheService));
     private const string SEARCH_SERVICES_CACHE_KEY = "search_services";
+    private const string SEARCH_CLIENTS_CACHE_KEY_PREFIX = "search_clients_";
     private static readonly TimeSpan CACHE_DURATION_SERVICES = TimeSpan.FromHours(1);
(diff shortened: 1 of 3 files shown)


jongio:
@hallipr 

Instructions to reproduce:
* Create Azure Search service (used Free SKU, you get 1 of these per subscription, but if you need more ping me with the sub id and I'll get you more)
* Enable system-assigned managed identity to that search service
* Create Azure OpenAI resource
* Create new model deployment in Azure OpenAI, for the model "text-embedding-003-small"
* In the OpenAI resource, add RBAC "Cognitive Services OpenAI Contributor" role for Azure Search service identity
* In the storage account you're using, add RBAC "Storage Blob Data Reader" role for the Azure Search service identity
* Create storage container "searchdocs", added a few docs (using Seth's "contoso outdoors" product description markdown files), and set the "category" metadata field of each blob randomly to A, B or C
* Created an Azure Search index. You can get the full index definition in JSON from the portal and use exactly that.
* Setup indexer/datasource/etc. Again you can get these in JSON form from the portal and then create them directly. Order: index, datasource, skillset, indexer.
* Indexer will run on creation and populate the index. You won't need to run it again.

Accounts needs "search index data reader" and "Search data contributor" roles



```
    "SearchServiceName": "jongsearch",
    "SearchIndexName": "products"
}
```

pablocastro:
@hallipr one more step I forgot in the list @jongio shared: you need to give the identity the tests run under RBAC access to your Azure Search instance. For that you'll need to enable RBAC or keys+RBAC auth (defaults to keys only), and then add Search Index Data Reader and Search Service Contributor roles.

jongio:
@hallipr - Can you add the search service to the test env bicep?

hallipr:
> @hallipr - Can you add the search service to the test env bicep?
Seems like this will be a bit of multi-resource bicep orchestration with a post-deploy script.

hallipr (approved):

hallipr:
See #157 for bicep work

PR #97, opened 2025-05-03
Move git-hooks to eng/scripts and put the logic into powershell

hallipr:
- convert pre-commit to pre-push


eng/scripts/Install-GitHooks.ps1
```diff
@@ -0,0 +1,45 @@
...
+
+# Copy hook files
+Write-Host "Copying git hooks from $gitHooksSourceDir to $gitHooksTargetDir..."
+$hookFiles = Get-ChildItem -Path $gitHooksSourceDir -File
+
+foreach ($file in $hookFiles) {
+    $targetFile = Join-Path $gitHooksTargetDir $file.Name
+
+    Write-Host "  Copying $($file.Name)..."
+    Copy-Item -Path $file.FullName -Destination $targetFile -Force
```
joshfree, line 34: Instead of the hard coded delete of pre-commit earlier maybe do a rename of x to x.old right before this line.

src/AzureMcp.csproj
```diff
@@ -73,26 +73,18 @@
...
 
-    <!-- on mac/linux, make sure they're executable -->
-    <Exec Condition="Exists('$(MSBuildProjectDirectory)/../.git/hooks/pre-commit') and '$(OS)' != 'Windows_NT'"
-          Command="chmod +x &quot;$(MSBuildProjectDirectory)/../.git/hooks/pre-commit&quot;" />
-
-    <!-- on Windows, make sure they're executable -->
-    <Exec Condition="Exists('$(MSBuildProjectDirectory)/../.git/hooks/pre-commit') and '$(OS)' == 'Windows_NT'"
-          Command="attrib +x &quot;$(MSBuildProjectDirectory)/../.git/hooks/pre-commit&quot;" />
+    <!-- Display informative message if pwsh fails -->
+    <Warning Text="Could not install Git hooks. PowerShell Core (pwsh) is required to install Git hooks."
```
joshfree, line 87: Make sure this is also in the contributing.md instructions with a link to an install point

joshfree (approved):
Left some comments to follow up on

eng/scripts/Install-GitHooks.ps1
hallipr: I'll need both.   We're not adding a new pre-commit hook, but we are wanting to remove the existing one

hallipr: I added the back to .old as well as special casing the pre-commit hook delete if it matches pre-push

src/AzureMcp.csproj
hallipr: I added instructions to contributing.md and an install link to the warning message

joshfree (approved):
Love it. Thanks for making the updates

eng/scripts/Install-GitHooks.ps1
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env pwsh
```
weshaggard, line 1: I highly recommend against using git hooks as they generally do nothing but cause issues. I'm fine if folks want to play with them but we should not require or depend on them because we often have to disable them.

PR #100, opened 2025-05-04
Small fixes to powershell scripts

hallipr:
- fix paths to packages in Build-Local and Test-ModulePerformance
- remove windows `attrib +x` from Install-GitHooks
  - attrib `x` in windows is data scrub, not executable
  - windows executes files by extension, not attribute


eng/scripts/Install-GitHooks.ps1
```diff
@@ -46,10 +46,8 @@ foreach ($file in $hookFiles) {
     Copy-Item -Path $file.FullName -Destination $targetFile -Force
```
weshaggard, line 46: @hallipr remind me to get more info about the usages of githooks here as I'm generally not a fan of them.

PR #104, opened 2025-05-05
Implement Azure service principal authentication in AzCommand

jdubois:
This is a port of the https://github.com/jdubois/azure-cli-mcp project, coming from
https://github.com/jdubois/azure-cli-mcp/blob/76b00cab0c2526f743af41323e6cc832abc9fa23/src/main/java/io/github/jdubois/azureclimcp/AzureCliService.java#L54

As requested by @jongio


src/Commands/Extension/AzCommand.cs
```diff
@@ -105,6 +106,54 @@ protected override AzArguments BindArguments(ParseResult parseResult)
         return null;
     }
 
+    private static async Task<bool> AuthenticateWithAzureCredentialsAsync(IExternalProcessService processService, ILogger logger)
```
jongio, line 109: @christothes - Does Azure Identity support AZURE_CREDENTIALS?  I know we have moved away from this, but in some cases this is the only option for devs - container dev.    Or should we switch to asking them to use AZURE_CLIENT_ID, TENANT_ID, CLIENT_SECRET instead of AZURE_CREDENTIALS?

christothes: I'm not aware of any support we have for AZURE_CREDENTIALS

christothes: If we do decided to use AZURE_CLIENT_ID, TENANT_ID, CLIENT_SECRET, we could add EnvironmentCredential to the chain and it would get picked up automatically.

christothes: It looks like AZURE_CREDENTIALS is something started in GitHub Actions. Is there a reason we want to support that pattern here?

jdubois: The goal is to have people authenticate to Azure CLI using it, like in the Azure CLI MCP documentation at https://github.com/jdubois/azure-cli-mcp?tab=readme-ov-file#install-and-configure-the-server-with-docker (on which this was copied)

rohit-ganguly:
@g2vinay please review!

src/Commands/Extension/AzCommand.cs
jongio: @jdubois - Can you create an AzureCredentialsEnvironmentCredential that derives from TokenCredential in Azure.Idenitty?

This should be a good starting point: https://github.com/Azure/azure-sdk-for-net/blob/8e101ec21e839ddb846b95500c214013842396dd/sdk/identity/Azure.Identity/src/Credentials/EnvironmentCredential.cs#L51

g2vinay: @christothes 
Looking at impl in PR.
Details from AZURE_CREDENTIALS are used to auth as service principal via az cli.

Can we add this support to az cli credential to consume AZURE_CREDENTIALS file and auth as service principal ?

christothes: Is there a file path or is the content of the json set as the value of the AZURE_CREDENTIALS env var? I think for now we should wrap the below impl in a local instance of TokenCredential. We should consider if this should be added to the existing AzureCliCredential separately.

jongio: I agree, let's get it going in the MCP repo as a custom token credential and then see what the identity team thinks.

g2vinay: Okay, I will open a PR to add this as a custom token credential to the MCP repo.

g2vinay: synced with @christothes, he will open a PR to add this in MCP repo.

christothes: @jongio I started to look at this more closely and I'm not sure we need an implementation of `TokenCredential`. The reason is that it seems that we just want to ensure that `az` is logged in with the credentials in AZURE_CREDENTIALS but we don't need the actual token. If that is true, what we have here seems appropriate.

g2vinay: Once we introduce the abstraction of Cached Token Credential to optimize and provide in memory caching. 
This layer won't be able to inherit it. 
Wrapping it in Token Credential also allows to manage auth centrally in MCP services.

g2vinay: synced with @christothes 
There is no need for TokenCredential abstraction here. 
We just need to abstract out the impl into AzureCLILoginProvider under auth package.

jongio: Sounds good.  I'll let @g2vinay and @christothes work out a good plan

jongio:
@christothes took this as a fresh PR here https://github.com/Azure/azure-mcp/pull/104

jdubois:
@jongio I believe you meant https://github.com/Azure/azure-mcp/pull/204

Normally you should have write access to my PR, so you can contribute to it instead of creating a new one.

If that's not the case, maybe because of some security policy in our org: just ask me and I invite people to my fork. That way you would get write access to the branch and can modify directly the PR.

johnpapa:
> @christothes took this as a fresh PR here #104

I’m curious why a new PR was created … could any change you need be worked into this one, so we maintain the history and committers>

christothes:
> > @christothes took this as a fresh PR here #104
> 
> I’m curious why a new PR was created … could any change you need be worked into this one, so we maintain the history and committers>

Sorry - this was my fault. I'm happy to revert mine and contribute to this fork if I can get permissions.

jdubois:
@christothes as #204 is already merged, this isn't really worth the trouble.
Then can check if you already have write to the branch here? I believe you should have access by default, then maybe there's a security limitation because that's managed by our organization, and that would be interesting to know and solve

christothes:
> @christothes as #204 is already merged, this isn't really worth the trouble. Then can check if you already have write to the branch here? I believe you should have access by default, then maybe there's a security limitation because that's managed by our organization, and that would be interesting to know and solve

I checked and I can't push to this fork from mine. I don't think anyone gets push access to your forks by default. 

It appears that there are two ways to do this - 
- grant access to let someone push directly to your fork
- create a pull request from my fork to yours

Issue #102, opened 2025-05-04
CosmosDB commands - client caching issue


alexdrenea:
When using any Cosmos commands, the CosmosService caches the CosmosClient object ( see [here](https://github.com/Azure/azure-mcp/blob/b9ab8ff55b17c6f754398e9b160083a632e0d5d5/src/Services/Azure/Cosmos/CosmosService.cs#L86) ).

This causes the server to always try the same connection even though the user, though the client, asks to change the context to a different account.

joshfree:
Thanks for reporting this issue, @alexdrenea. @jongio can you help route this to the right owner?

alexdrenea:
I can happily take this over if needed

joshfree:
>I can happily take this over if needed 
much appreciated, @alexdrenea!  I tagged @xiangyan99 too, as a code review buddy so you have a point of contact

xiangyan99:
Fixed in https://github.com/Azure/azure-mcp/pull/231

PR #116, opened 2025-05-06
Cache CosmosClient by account name

alexdrenea:
This PR adds Caching for CosmosClient objects based on the requested account name (#102)

Also fixes a minor bug by validating the CosmosClient object upon creation so the auth key method actually gets re-triggered when credentials cannot be found.

### DIFF
--- a/src/Services/Azure/Cosmos/CosmosService.cs
+++ b/src/Services/Azure/Cosmos/CosmosService.cs
@@ -14,7 +14,7 @@ public class CosmosService(ISubscriptionService subscriptionService, ITenantServ
     : BaseAzureService(tenantService), ICosmosService, IDisposable
 {
     private const string CosmosBaseUri = "https://{0}.documents.azure.com:443/";
-    private CosmosClient? _cosmosClient;
+    private Dictionary<string, CosmosClient> _cosmosClientCache = [];
     private bool _disposed;
     private readonly ISubscriptionService _subscriptionService = subscriptionService ?? throw new ArgumentNullException(nameof(subscriptionService));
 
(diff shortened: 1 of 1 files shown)


src/Services/Azure/Cosmos/CosmosService.cs
```diff
@@ -14,7 +14,7 @@ public class CosmosService(ISubscriptionService subscriptionService, ITenantServ
     : BaseAzureService(tenantService), ICosmosService, IDisposable
 {
     private const string CosmosBaseUri = "https://{0}.documents.azure.com:443/";
-    private CosmosClient? _cosmosClient;
+    private Dictionary<string, CosmosClient> _cosmosClientCache = [];
```
jongio, line 17: Please use the existing CacheService.  See how Search is doing it.  Thanks

jongio (changes requested):

src/Services/Azure/Cosmos/CosmosService.cs
jongio: https://github.com/Azure/azure-mcp/pull/90/files has that code

alexdrenea: Updated to follow pattern in Search - but I am not convinced this is the right approach.

The CosmosClient is meant to be a long lived object that should be around for the duration of the execution so tying it's lifecycle to the calling service seemed appropriate. 

When using the cache, we do not control what happens with the object when it gets evicted so we don't get an opportunity to properly dispose. 
If we set a very long expiration (essentially never evict) then we have an issue trying to find all the clients in the cache so we can dispose them properly when the base CosmoService gets disposed. (see empty Dispose method now)

Looking for guidance on how to address this!
Thanks

xiangyan99: Could you provide more details about
"we have an issue trying to find all the clients in the cache so we can dispose them properly when the base CosmosService gets disposed."?

In practical situations, shouldn't there be a limited number of cached clients?

alexdrenea: Since the user can ask the about multiple Cosmos Accounts, we will have a CosmosClient instance for each account the user is asking about. We will cache every CosmosClient based on the account name ("cosmos_clients_account1", "cosmos_clients_account2" , etc..)

When we cache these clients locally in a dictionary, it's easy to dispose them all since we easily query the dictionary. 
In the absence of a method that can return list of objects based on key prefix on ICacheClient it will be difficult to find all Clients in the Cache in order to dispose them.

xiangyan99: I think this is a valid request.

Let's go ahead to merge the fix for //build and I opened https://github.com/Azure/azure-mcp/issues/139 to track the request.

xiangyan99 (approved):

src/Services/Azure/Cosmos/CosmosService.cs
jongio: Let's do this in a generic way so that other clients can benefit.  If the CacheService doesn't meet needs, then how can we update to meet them?

xiangyan99:
@alexdrenea we updated the interface of cache service to be

public ValueTask SetAsync<T>(string group, string key, T data, TimeSpan? expiration = null)

Does this meet your requirements?

If yes, could you rebase and update your PR?

xiangyan99:
Fixed in https://github.com/Azure/azure-mcp/pull/231

PR #168, opened 2025-05-12
Introducing cache group for the scenario that one service needs to ca…

xiangyan99:


src/Services/Azure/ResourceGroup/ResourceGroupService.cs
```diff
@@ -13,6 +13,7 @@ public class ResourceGroupService(ICacheService cacheService, ISubscriptionServi
 {
     private readonly ICacheService _cacheService = cacheService ?? throw new ArgumentNullException(nameof(cacheService));
     private readonly ISubscriptionService _subscriptionService = subscriptionService ?? throw new ArgumentNullException(nameof(subscriptionService));
+    private const string CACHE_GROUP = "resourceGroup";
```
jongio, line 16:
```suggestion
    private const string CACHE_GROUP = "resourcegroup";
```

src/Services/Caching/CacheService.cs
```diff
@@ -9,29 +10,99 @@ namespace AzureMcp.Services.Caching;
 public class CacheService(IMemoryCache memoryCache) : ICacheService
 {
     private readonly IMemoryCache _memoryCache = memoryCache;
+    private static readonly ConcurrentDictionary<string, HashSet<string>> _groupKeys = new();
 
-    public ValueTask<T?> GetAsync<T>(string key, TimeSpan? expiration = null)
+    public ValueTask<T?> GetAsync<T>(string group, string key, TimeSpan? expiration = null)
```
jongio, line 15: group shold be optional right?

xiangyan99: No. It is required. It connects to the service. All clients using the same service should be in one group.

hallipr:
Is this to avoid cache collision between different service using the same basic key names for different content?

xiangyan99:
> Is this to avoid cache collision between different service using the same basic key names for different content? And do any services retrieve cache entries created by other services?

We also want to allow a service to get all the clients it has saved. 

In the case of Cosmos, it makes a client for each account. When the Cosmos service is shut down, it wants to close all the clients it created, so it needs a way to get all of them.

hallipr:
> We also want to allow a service to get all the clients it has saved.

It would seem like following the ILogger<T> pattern would be nice.    `ICacheService<T> cacheService`

Where ICacheService<T> is just ICacheService with name derived from T

https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerT.cs#L50

src/Services/Azure/Subscription/SubscriptionService.cs
```diff
@@ -12,6 +12,7 @@ public class SubscriptionService(ICacheService cacheService, ITenantService tena
     : BaseAzureService(tenantService), ISubscriptionService
 {
     private readonly ICacheService _cacheService = cacheService ?? throw new ArgumentNullException(nameof(cacheService));
+    private const string CACHE_GROUP = "subscription";
```
hallipr, line 15: The snake cased constants are a c# convetion break.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

I think we need another PR to fix all of them

hallipr: https://github.com/dotnet/runtime/blob/fdbe38ce16671ae356e2bad21479e38b61d03340/src/libraries/System.Private.CoreLib/src/System/Char.cs#L48

```cs
        public const char MinValue = (char)0x00;

        private const byte IsWhiteSpaceFlag = 0x80;
        private const byte IsUpperCaseLetterFlag = 0x40;
        private const byte IsLowerCaseLetterFlag = 0x20;
        private const byte UnicodeCategoryMask = 0x1F;
```

xiangyan99:
> cacheService

Seems like we need to reconsolidate the ICacheService interface. I will let Jon make the call.

src/Services/Azure/Subscription/SubscriptionService.cs
xiangyan99: Opened https://github.com/Azure/azure-mcp/issues/180 to track the issue.

src/Services/Caching/CacheService.cs
jongio: I guess I'm not following why group would be required for all cache entries.

jongio: Okay, it's fine as-is

hallipr (approved):

PR #179, opened 2025-05-12
Implement caching system and refactor subscription handling

jongio:
Introduce a caching system with memory and file providers, while simplifying command argument methods and updating subscription handling to use new data structures. Add tests to ensure cache functionality and fix formatting issues in related tests.


src/Services/Caching/Providers/MemoryCacheProvider.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+using System.Threading.Tasks;
+using AzureMcp.Services.Caching.Shared;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace AzureMcp.Services.Caching.Providers;
+
+public class MemoryCacheProvider(IMemoryCache memoryCache) : ICacheProvider
+{
+    public ValueTask<T?> GetAsync<T>(string key)
```
hallipr, line 10: Why are we hiding the cache miss behind what looks like a cache hit for null?

src/Services/Caching/Providers/MemoryCacheProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+        => memoryCache.TryGetValue(key, out T? value) ? new ValueTask<T?>(value) : default;
```
hallipr, line 11: Why are we hiding the cache miss behind what looks like a cache hit for null?

jongio:
Going to do this differently

PR #189, opened 2025-05-13
Add support for browser auth timeout

g2vinay:
Adds support for browser auth opt-in timeout.

Handles cases where user cancels browser auth by closing the browser, but the process keeps waiting on a response indefinitely. 
Opt-in behavior adds timeout support for this case.

P.S, I expect this support to go away/not be required, once WAM broker support lands in Mac and Linux.


src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -89,7 +90,16 @@ private static InteractiveBrowserCredential CreateBrowserCredential(string? tena
             brokerOptions.ClientId = clientId;
         }
 
-        return new(brokerOptions);
+        var browserCredential = new InteractiveBrowserCredential(brokerOptions);
+
+        // Check for timeout value in the environment variable
+        string? timeoutValue = Environment.GetEnvironmentVariable(BrowserAuhtenticationTimeoutEnvVarName);
```
JasonYeMSFT, line 96: Does it make sense to always have a timeout with some default value, like 5 minutes? If someone uses your MCP and hits this, it would be hard to recover because they will need to search for the process and kill it.

g2vinay: 5 minute timeout as default, sounds fine to me.
we can adjust it, according to user feedback, as it comes.
the env var is still there to change it.

joshfree (approved):

joshfree (approved):

PR #247, opened 2025-05-23
It is by design that we cannot get tenant name in TME.

xiangyan99:
Disable the live test that needs it.


philon-msft (approved):

xiangyan99:
The deployment issue is known and addressed in another PR.

Issue #250, opened 2025-05-24
Azure MCP Server fails to query Application Insights tables with JSON parsing error


chris-battle-bunnings:
# Azure MCP Server fails to query Application Insights tables with JSON parsing error

## Bug Description

The Azure MCP Server's `azmcp-monitor-log-query` tool consistently fails when querying Application Insights tables (e.g., `AppTraces`) with a JSON parsing error, while the same queries work perfectly in the Azure Portal and other Azure Log Analytics tables work fine through the MCP server.

## Error Details

**Primary Error:**
```
Error querying logs: Error querying workspace: 'e' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
```

**Secondary Error (when using legacy table names):**
```
BadArgumentError, SemanticError - "'order' operator: Failed to resolve table or column expression named 'traces'" (SEM0100)
```

## Environment Information

- **Azure MCP Server Version:** `@azure/mcp@latest` (installed via npx)
- **Client:** Cursor IDE with MCP support
- **Workspace:** Workspace-based Application Insights resource
- **Workspace ID:** `ae51d3eb-6ebc-4867-84a9-dd29aa5fd64f-YouTubeWatcherRG-EAU`
- **Subscription:** `ae51d3eb-6ebc-4867-84a9-dd29aa5fd64f`
- **Resource Group:** `YouTubeWatcherRG`
- **Authentication:** Azure CLI (`az login`) - confirmed working

### What Works ✅

1. **Standard Log Analytics tables:**
   ```
   Tool: azmcp-monitor-log-query
   Parameters:
   - table-name: "FunctionAppLogs" 
   - query: "FunctionAppLogs | order by TimeGenerated desc | take 15"
   Result: Success (returns empty array, but no errors)
   ```

2. **Azure Portal queries:**
   - Same Application Insights workspace
   - Query: `traces | order by timestamp desc | take 15`
   - Result: Returns data successfully

### What Fails ❌

3. **Application Insights tables via MCP:**
   ```
   Tool: azmcp-monitor-log-query
   Parameters:
   - table-name: "AppTraces"
   - query: "AppTraces | order by TimeGenerated desc | take 15"
   Result: Error querying logs: Error querying workspace: 'e' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
   ```

4. **Legacy table names:**
   ```
   Tool: azmcp-monitor-log-query  
   Parameters:
   - table-name: "traces"
[...]

Issue #244, opened 2025-05-22
Exception when querying log analytics workspace


suyadav1:
While querying the Log Analytics Workspace, the agent consistently throws an exception. The server is able to successfully execute other operations, such as retrieving all workspaces in the subscription and listing all tables within a workspace. The issue occurs only when attempting to query a table.

The error returned is:

`"message":"Error querying logs: Error querying workspace: \u0027a\u0027 is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.`

This is a **blocker**, so please prioritize and address it urgently.

Attached is a screenshot of the agent’s input and output. 
Note: the issue occurs regardless of whether we use the workspace name or ID.

![Image](https://github.com/user-attachments/assets/b5884ec1-fe78-4cca-a45c-0c399b4fab69)

loneshaana:
There is an open issue for the same bug , this seems to be duplicate https://github.com/Azure/azure-mcp/issues/240

jongio:
I can repro this, investigating...

Issue #240, opened 2025-05-22
JsonNode.Parse throws exception when parsing non json-formatted values in QueryWorkspace function of MonitorService.cs


loneshaana:
The code uses JsonNode.Parse(row[i]?.ToString() ?? "null") to convert values returned from QueryWorkspaceAsync() into JsonNode objects. However, this approach is flawed when row[i] is a non-JSON-formatted string (e.g., plain text, timestamps, or any .NET object with ToString() output that doesn't conform to JSON syntax).

<img width="1906" alt="Image" src="https://github.com/user-attachments/assets/2868caac-746e-4a95-803f-d94e8dd230df" />

## Steps to  Reproduce
1. Query a workspace that returns non json values Example:
```kusto
AppTraces
| take 1
| project Message,TimeGenerated
```
3. Observe the exception execution 

## Proposed Fix
Replace the problematic line
```csharp
rowDict[columns[i].Name] = JsonNode.Parse(row[i]?.ToString() ?? "null");
```
with a safe and type-aware alternative
```csharp
rowDict[columns[i].Name] = JsonValue.Create(row[i]);
```

loneshaana:
A quick way of reproducing this would be to run the below command

npx @azure/mcp@latest monitor log query --tenant '<TENANT_ID>' --table-name 'AppTraces' --workspace '<WORKSPACE_ID>' --resource-group '<RESOURCE_GROUP>' --query 'AppTraces | take 1 | project Message' --subscription '<SUBSCRIPTION_ID>'

PR #280, opened 2025-05-31
Fix Monitor Logs Query and Add Tests

jongio:
Enhance code organization and logging capabilities by refactoring command classes, implementing cancellation tokens, and improving error handling in log sending methods. Remove unnecessary methods and streamline authentication processes.

Fixes https://github.com/Azure/azure-mcp/issues/250, Fixes https://github.com/Azure/azure-mcp/issues/244, Fixes https://github.com/Azure/azure-mcp/issues/240


src/Services/Azure/Monitor/MonitorService.cs
```diff
@@ -197,21 +196,44 @@ public async Task<List<JsonNode>> QueryLogs(
...
+            var options = AddDefaultPolicies(new LogsQueryClientOptions());
+
+            if (retryPolicy != null)
+            {
+                options.Retry.Delay = TimeSpan.FromSeconds(retryPolicy.DelaySeconds);
+                options.Retry.MaxDelay = TimeSpan.FromSeconds(retryPolicy.MaxDelaySeconds);
+                options.Retry.MaxRetries = retryPolicy.MaxRetries;
+                options.Retry.Mode = retryPolicy.Mode;
+                options.Retry.NetworkTimeout = TimeSpan.FromSeconds(retryPolicy.NetworkTimeoutSeconds);
+            }
```
srnagar, line 209: If retry policy is set, are all these properties required to be set as well? If not, should we continue to keep the defaults and update only non-null values? For e.g. the user may just want to update the `MaxDelay` and not set the `Delay` in retry policy options.

tests/AzureMcp.Tests.csproj
```diff
@@ -12,6 +12,7 @@
   </PropertyGroup>
   
   <ItemGroup>
+    <PackageReference Include="Azure.Monitor.Ingestion" />
```
srnagar, line 15: Just curious if we need to use the ingestion package to send logs or is there a way to send the logs through OTel?

tests/Helpers/LogAnalyticsHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+    private async Task<string> GetWorkspaceIdAsync()
+    {
+        if (!string.IsNullOrEmpty(_workspaceId))
+        {
+            return _workspaceId;
+        }
+
+        // Get workspace info using the monitor service
+        var workspaces = await _monitorService.ListWorkspaces(_subscription, _tenantId);
+        var workspace = workspaces.FirstOrDefault(w => w.Name.Equals(_workspaceName, StringComparison.OrdinalIgnoreCase))
```
srnagar, line 50: Do we expect to have just 1 workspace returned from the `ListWorkspaces` call?

tests/Helpers/LogAnalyticsHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+            var endpoint = new Uri($"https://{customerId}.ods.opinsights.azure.com");
+            var options = new LogsIngestionClientOptions
+            {
+                Retry =
+                {
+                    MaxRetries = 3,
+                    Delay = TimeSpan.FromSeconds(2),
+                    MaxDelay = TimeSpan.FromSeconds(10)
+                }
+            };
```
srnagar, line 87: nit: Do we need a custom retry configuration? It would be good to add a comment why the default retry configuration won't work.

tests/Helpers/LogAnalyticsHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+    {
+        var client = await GetLogsIngestionClientAsync(customerId).ConfigureAwait(false);
+
+        try
+        {
+            using var content = RequestContent.Create(logs);
+            _logger.LogInformation("Sending {Count} logs to workspace {WorkspaceId}", logs.Length, customerId);
+
+            cancellationToken.ThrowIfCancellationRequested();
+            var response = await client.UploadAsync(
```
srnagar, line 170: Will this return a response even if the service returned a non-2xx status or does it throw an exception here?

srnagar (approved):

tests/Helpers/LogAnalyticsHelper.cs
jongio: Yes

tests/Helpers/LogAnalyticsHelper.cs
jongio: Is there a question behind the question?  I'm not sure what you are getting at.

tests/Helpers/LogAnalyticsHelper.cs
jongio: That is what I think is best for the test.

tests/AzureMcp.Tests.csproj
jongio: I'm not sure, do you know?

src/Services/Azure/Monitor/MonitorService.cs
jongio: What do you recommend?

tests/Helpers/LogAnalyticsHelper.cs
srnagar: In the `catch` block, we always return `HttpStatusCode.InternalServerError` for any exception. So, I wasn't sure what the behavior of `client.UploadAsync()` call would be if the service return other error codes like 401 or 403. Would that return a response or would that throw an exception? If it returns the response, then this looks good. If it throws an exception, we will be mapping all service errors to `InternalServerError`.

tests/AzureMcp.Tests.csproj
srnagar: tagging @lmolkova to see if we can use .NET OTel lib to send the logs to Azure Monitor.

tests/Helpers/LogAnalyticsHelper.cs
jongio: Given this is a just a test helper I don't think it matters. Do you?

src/Services/Azure/Monitor/MonitorService.cs
jongio: Since these are all value types and it's just setting to the default, I think any check like that would be a breaking change to make them nullable or not be worth the effort.

srnagar: Makes sense, thanks for checking.

tests/Helpers/LogAnalyticsHelper.cs
srnagar: For testing purposes, it shouldn't matter if we don't need to assert the status codes. Resolving the comment.

tests/AzureMcp.Tests.csproj
lmolkova: just keep using ILogger and just enable otel - it'll forward all ilogger to AzMon, no need for low-level Azure.Monitor.Ingestion

jongio: We went with the Ingestion APIs because we need a custom table, which the Logger/Otel approach doesn't seem to support. Since this is working as-is and the suggestion is an improvement, I'm going to create this issue and get this fix in.

https://github.com/Azure/azure-mcp/issues/290

PR #285, opened 2025-06-02
Adding support for --debug and developer documentation to debug live test

anuchandy:
* Added a --debug option to the azmcp server.
* Updated the live test process to utilize the new --debug flag.
* added developer documentation with instructions on how to debug a live test.

### DIFF
--- a/tests/Client/Helpers/LiveTestFixture.cs
+++ b/tests/Client/Helpers/LiveTestFixture.cs
@@ -30,22 +30,28 @@ public class LiveTestFixture : LiveTestSettingsFixture
         {
             Environment.CurrentDirectory = Settings.SettingsDirectory;
             transportOptions.Command = "npx";
-            if (enableDebugMode) {
+            if (enableDebugMode)
+            {
                 // start the server with --debug option to wait for it to attach.
                 transportOptions.Arguments = ["-y", Settings.TestPackage, "server", "start", "--debug"];
-            } else {
+            }
+            else
+            {
                 transportOptions.Arguments = ["-y", Settings.TestPackage, "server", "start"];
             }
         }
 
         var clientTransport = new StdioClientTransport(transportOptions);
-        if (enableDebugMode) {
+        if (enableDebugMode)
+        {
             var clientOptions = new McpClientOptions
             {
                 InitializationTimeout = TimeSpan.FromMinutes(2)
             };
             Client = await McpClientFactory.CreateAsync(clientTransport, clientOptions);
-        } else {
+        }
+        else
+        {
             Client = await McpClientFactory.CreateAsync(clientTransport);
         }
     }


docs/dev/Debug-LiveTest.md
```diff
@@ -0,0 +1,123 @@
...
+        while (!System.Diagnostics.Debugger.IsAttached)
+        {
+            Thread.Sleep(100);
+        }
+        LogDebug("[ServiceStartCommand] Debugger Attached");
+    }
+#else
+    LogDebug("[ServiceStartCommand] ExecuteAsync called in non-debug mode");
+#endif
+...
```
anuchandy, line 101: Discussed with @joshfree , we can make the debugging experience better by having all these code in azmcp, which get's activated under `#Debug` and `-debug` flag. I'll do that in a followup pr.

joshfree (approved):

CONTRIBUTING.md
```diff
@@ -138,6 +138,8 @@ This will produce .tgz files in the `.dist` directory and set the `TestPackage`
   "TestPackage": "file://D:\\repos\\azure-mcp\\.dist\\wrapper\\azure-mcp-0.0.12-alpha.1746488279.tgz"
```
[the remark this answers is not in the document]
anuchandy: Hi @jongio, @hallipr, I have updated the pull request with the following changes:

* the build scripts (Build-Module.ps1, Build-Local.ps1) now accept the `-DebugBuild` flag. When this flag is set, the package is built using `/p:Configuration=Debug`.
* Azure MCP Server can now be started with the `--debug` flag. If this flag is set, the server will wait for the debugger to attach in debug mode.

Let me know your thoughts. I'll also update the `Debug-LiveTest.md` file accordingly.

docs/dev/Debug-LiveTest.md
anuchandy: resolving this, as the code change is now included in the pr.

tests/Client/Helpers/LiveTestFixture.cs
```diff
@@ -28,11 +28,22 @@ public override async ValueTask InitializeAsync()
         {
             Environment.CurrentDirectory = Settings.SettingsDirectory;
             transportOptions.Command = "npx";
+#if DEBUG
```
hallipr, line 31: I'm not sure how useful the DEBUG flag is in a test project.   Projects default to debug and we haven't chanted that default, so I would assume that all of our tests are run in debug mode unless someone manually changes to Release before running tests.

But the test's build configuration is orthogonal to the build configuration used when building the package or wanting to be able to attach a debugger at server start.

I think this fits better as a test setting.  i.e.  There's no reason to worry about what gets compiled into the test dll, so I'd avoid compilation conditionals and just rely on the TestSettings poco and json file

hallipr:
I'm starting to question the necessity of the feature.
- We only need to wait for a debugger when testing the server in TestPackage mode (through an npx call).
- We typically would only run in TestPackage mode locally to detect parameter passing issues (or maybe some odd stdio stream binding issue), but `server start` takes no parameters.  It's just much faster to change code and run the test immediately than to wait for a local package to be built.
- We do run in DEBUG and TestPackage mode for all CI runs, but there's never a debugger attached
- For all other local runs, there will already be a debugger attached and you can just set a breakpoint in server start if needed. (we should probably still add the InitializationTimeout to accommodate stepping in the start command)

anuchandy:
> I like the part about building the package in debug mode.
> 
> I'm not sure we need to wait on debugger attach in `server start` unless we're seeing issues with that command only when in an npx call. For all other tests, we can set a breakpoint in the test, attach to the server process, then continue the test.

Thank you, Patrik @hallipr. Here is what failed for me that lead to this pr,

1.	I intended to debug one of the tests located under "tests\Client" and step into a file within "src\Command".
2.	So I set a breakpoint in one of the command files in "src\Command".
3.	I right clicked on the test (that invokes CallToolAsync) and selected “Start Debugging.”
4.	However, the breakpoint set in step 2 did not get hit.

Based on my understanding, the issue arises because, when running tests in "tests\Client", CallToolAsync must use the studio and communicate with a different process i.e, the client (test) and server (azmcp) runs in separate processes.

Probably I overlooked as I'm new to .net, but I couldn’t make this debugging flow work without having that seperate azmcp server process to wait for attach.

jongio:
I'd say work with @hallipr to get this one addressed and merged.

tests/Client/Helpers/LiveTestFixture.cs
```diff
@@ -28,11 +28,22 @@ public override async ValueTask InitializeAsync()
...
+            transportOptions.Arguments = ["-y", Settings.TestPackage, "server", "start", "--debug"];
+#else
             transportOptions.Arguments = ["-y", Settings.TestPackage, "server", "start"];
+#endif
         }
 
         var clientTransport = new StdioClientTransport(transportOptions);
-
+#if DEBUG
```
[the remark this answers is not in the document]
anuchandy: updated to wait for 2 minutes, should be enough for dev to find the process id to be attached.

tests/Client/Helpers/LiveTestFixture.cs
anuchandy: great point, I moved it to the LiveTestSettings.cs, so one can turn this on-off using `.testsettings.json`

PR #298, opened 2025-06-03
Temporarily skip monitor logs query test

jongio:
Skip the monitor logs query test to address existing issues.


tests/Client/MonitorCommandTests.cs
```diff
@@ -76,7 +76,7 @@ public async Task Should_list_monitor_workspaces()
         Assert.NotEmpty(array);
     }
 
-    [Fact()]
+    [Fact(Skip = "Temporary skip to fix the test")]
```
joshfree, line 79: would be nice if this included the tracking bug number to fix/reenable

PR #333, opened 2025-06-09
Add Azure SQL Advise Command

jongio:
Adding first Azure SQL command that will be useful for developers wanting to improve their SQL implementation.

User Scenario: "help me optimize my azure sql db"

This tool will return optimization recommendations from sql advisor.


.vscode/launch.json
```diff
@@ -415,12 +415,42 @@
             "env": {
                 "DOTNET_ENVIRONMENT": "Development"
             }
+        },        {
```
xiangyan99, line 418: New line?

src/Commands/Server/ToolOperations.cs
```diff
@@ -75,6 +74,22 @@ private async ValueTask<CallToolResponse> OnCallTools(RequestContext<CallToolReq
...
+
+            _logger.LogWarning(content.Text);
+
+            return new CallToolResponse
+            {
+                Content = [content],
+                IsError = true,
+            };
+        }
+
```
wbreza, line 92: I addressed this in another PR.  You can remove this.

xiangyan99:
Please also update docs/azmcp-commands.md.

src/Commands/Sql/Db/DbAdviseCommand.cs
```diff
@@ -0,0 +1,140 @@
...
+        - Get AI-powered insights for database tuning
+        - Identify and fix performance regressions
+        
+        Required options:
+        - database: The name of the database to analyze and optimize
+        - server-name: The name of the SQL server containing the database
+        Optional:
+        - table: Focus optimization recommendations on a specific table
+        - minimum-impact: Set minimum performance impact threshold for recommendations
+        - advisor-type: Filter by specific advisor type (CreateIndex, DropIndex, ForceLastGoodPlan, DbParameterization)
```
wbreza, line 49: Isn't this redundant since this information should already be present in the tool argument descriptions?

src/Models/Sql/SqlRecommendation.cs
```diff
@@ -0,0 +1,72 @@
...
+// Licensed under the MIT License.
+
+using System.Text.Json.Serialization;
+
+namespace AzureMcp.Models.Sql;
+
+/// <summary>
+/// Represents a recommendation for a SQL database from Azure SQL Database advisors.
+/// </summary>
+public sealed record SqlRecommendation
```
wbreza, line 11: Why do we need to re-expose return/result types vs leveraging ones directly from the SDKs? We're returning JSON anyway - so why do we need all these boilerplate types?

src/Services/Interfaces/ISqlService.cs
```diff
@@ -0,0 +1,64 @@
...
+    Task<SqlAnalysisResult> GetRecommendationsAsync(
+        string database,
+        string server,
+        string resourceGroup,
+        string? tableName,
+        int? minImpact,
+        string? advisorType,
+        string subscription,
+        string? tenant = null,
+        RetryPolicyOptions? retryPolicy = null);
```
wbreza, line 63: nit: Consider the group of optional properties in some options object.

wbreza (approved):
Looks good overall - just a few low priority comments.  Main comment is around whether or not we need to re-expose types for the command results vs leveraging existing objects from the SDKs.  Are we trimming down the amount of data or other reasons?

src/Services/Interfaces/ISqlService.cs
jongio: I think we should think about that holistically since this is the pattern to date.

src/Commands/Server/ToolOperations.cs
jongio: Undid changes for this file.

src/Commands/Sql/Db/DbAdviseCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+            HandleException(context.Response, ex);
+        }
+
+        return context.Response;
+    }
+    protected override string GetErrorMessage(Exception ex) => ex switch
+    {
+        DatabaseNotFoundException => "Database not found. Verify the database exists and you have access.",
+        _ => base.GetErrorMessage(ex)
+    }; internal record DbAdviseCommandResult(SqlAnalysisResult Analysis) : IDbAdviseCommandResult
```
anuchandy, line 126: nit; should this record definition be in a new line?

src/Services/Azure/Sql/SqlService.cs
```diff
@@ -0,0 +1,402 @@
...
+    {
+        if (string.IsNullOrEmpty(sql))
+            return string.Empty;
+
+        // Extract table name from CREATE INDEX statements
+        // Pattern: CREATE [NONCLUSTERED] INDEX ... ON [schema].[table] or ON [table]
+        var patterns = new[]
+        {
+            @"ON\s+\[?([^\]\s\[]+)\]?\.\[?([^\]\s\[]+)\]?",  // ON [schema].[table] or ON schema.table
+            @"ON\s+\[?([^\]\s\[]+)\]?(?!\s*\.)",             // ON [table] (without schema)
```
anuchandy, line 286: In Java, we try to compile the regex upfront so it can help with perf. 

Reading the doc, .NET seems to have similar concept https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators#compiled-regular-expressions, so we could avoid runtime regx compilation (and is AOT friendly).

```C#
    // / ON [schema].[table] or ON schema.table
    [GeneratedRegex(@"ON\s+\[?([^\]\s\[]+)\]?\.\[?([^\]\s\[]+)\]?",  RegexOptions.IgnoreCase)]
    private static partial Regex NameFromTableSchemaRegex();
    
   // // ON [table] (without schema)
    [GeneratedRegex(@"ON\s+\[?([^\]\s\[]+)\]?(?!\s*\.)",  RegexOptions.IgnoreCase)]
    private static partial Regex NameFromTableOnlyRegex();
```

```C#
var match = NameFromTableSchemaRegex().Match(sql);
// rest is the same code.
```

anuchandy (approved):
LGTM, ty!

jongio (commented):
I'm still working on getting an env that has db recommendations so I can test this.

jongio:
This is outdated. Will discuss options with SQL team.

PR #366, opened 2025-06-13
Proxy tool per service and initial support for remote servers

wbreza:
## What does this PR do?

In this PR we are experimenting with an approach where instead of exposing a single `azure` tool we expose a tool per service area like `storage` and `keyvault`.

Internally the same proxy approach and tool narrowing approach is applied but gives users more details on what tools/services are exposed which also comes with more user controls on allowing tools to be invoked.

This POC also includes a pathway to enable support for remote MCP servers.  The newly available Microsoft Learn is a remote MCP server and is mapped to and exposed as the "learn" tool from Azure MCP and child requests are forwarded to the remote MCP learn tool like any of our other included MCP tools.

image

## How to Enable & Run

`azmcp server start --service proxy`

This PR also introduces an embedded resource `registry.json` that supports registering other MCP servers that will be exposed through the Azure MCP.

```json
{
    "servers": {
        "learn": {
            "url": "https://learn.microsoft.com/api/mcp",
            "description": "Search official Microsoft/Azure documentation..."
        }
    }
}
```

## mcp.json

```json
"servers": {
  "Azure": {
    "type": "stdio",
    "command": "azmcp",
    "args": ["server", "start", "--service", "proxy"]
  }
}
```

## Next steps
- Update contributing & other docs to enable this support in a future PR.


jongio (commented):
Looks like a good update.

wbreza:
> > > Is it possible to modify registry.json after installation?
> > 
> > 
> > Not right now. I'm treating this as a controlled resource that we explicitly maintain as part of releases vs it being a user configurable setting.
> > The idea is that we will use this file to control official partner team MCP servers that we want to light up by default when Azure MCP is installed.
> 
> Let's make sure post-installation-modification is prohibited to prevent potential security risks.

It's an embedded resource in the dll/exe - so does not land on the users machine as a file to modify.  We should be good to go.

PR #372, opened 2025-06-16
docs: Add troubleshooting section for MCP 128-tool limitation

samvaity:
## What does this PR do?
Adds a prominent troubleshooting section for the "You may not include more than 128 tools in your request" error, which is becoming our most commonly reported issue.

## Changes
- 🚨 **New section at top of troubleshooting guide** highlighting the 128-tool limit
- 📝 **Two clear workarounds** with practical examples:
  - **Option 1** : Use VS Code Custom Chat Modes (Recommended)*
  - **Option 1**: Selective tool loading (multiple targeted servers)
  - **Option 2**: Dynamic tool selection (single `--service azure` proxy)
- 🔧 **Updated examples** using the correct `npx @azure/mcp@latest` command
- ⚠️ **Clear warnings** about when each approach helps vs. doesn't help

## Why this change?
Users frequently encounter this error when combining Azure MCP "all" + GitHub MCP "all" toolsets. This guide provides immediate, actionable solutions.

## Testing
- [x] Verified `--service azure` creates the documented dynamic proxy tool
- [x] Confirmed examples use correct npm package (`@azure/mcp` not `@azure/mcp-server`)
- [x] Validated configuration syntax matches existing troubleshooting examples

## Type of Change
- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change


TROUBLESHOOTING.md
```diff
@@ -1,5 +1,75 @@
 # Troubleshooting
 
+## ⚠️ Known Issue: "You may not include more than 128 tools in your request"
```
joshfree, line 3: this is a VS Code specific limit.  Other Agents have different limits (some even lower); could you tweak the title to mention VS Code?

TROUBLESHOOTING.md
```diff
@@ -1,5 +1,75 @@
...
+      "type": "stdio",
+      "command": "npx",
+      "args": ["-y", "@modelcontextprotocol/server-github@latest"]
+    }
+  }
+}
+```
+*Result: ~15-20 tools total instead of 128+*
+
+**Available Azure Services for `--service` flag:**
```
joshfree, line 44: could you add a link to the anchor in readme.md too

TROUBLESHOOTING.md
```diff
@@ -1,5 +1,75 @@
...
+⚠️ **Note**: This still counts as 1 tool toward the 128 limit, but that 1 tool can access all Azure services. However, if you're combining with GitHub MCP "all" or other comprehensive toolsets, you may still hit the 128-tool limit.
+
+### How to Check Your Tool Count
+To see how many tools you're loading:
+1. Open VS Code Command Palette (Ctrl+Shift+P)
+2. Run "MCP: List Servers"  
+3. Check the tool count for each server in the output window
+
+### Future Considerations
+We are exploring options for tool consolidation under shared namespaces or aliases to simplify usage while staying within the 128-tool limitation.
```
joshfree, line 71: VS Code CoPilot is planning to address the tool limit in a future update.  Their guidance to us is that it's a client problem, and something the MCP Servers shouldn't have to worry about.  Exact timeline and changes on their side is TBD.

joshfree (changes requested):
left a few nits

TROUBLESHOOTING.md
```diff
@@ -1,5 +1,75 @@
...
+
+### Problem
+When configuring Azure MCP with 'all' toolsets for convenience, you may encounter this error:
+
+![128 tools limit error](docs/images/128-tools-limit-error.png)
+
+The error message states: *"You may not include more than 128 tools in your request"*
+
+### Root Cause
+MCP clients have a limitation of 128 tools maximum per request. When you combine multiple comprehensive toolsets (like GitHub MCP 'all' + Azure MCP 'all'), the total number of available tools exceeds this limit.
```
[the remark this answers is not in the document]
charris-msft: but spell "Copilot" without a capital "P" :)

TROUBLESHOOTING.md
```diff
@@ -1,5 +1,75 @@
...
+
+### Workarounds
+
+**Option 1: Use Selective Tool Loading (Recommended)**
```
charris-msft, line 17: I may be missing something, but this seems like it would work well only if a dev knows they will only ever need a specific set of tools. Otherwise they will have to manually edit the list every time they want to add/remove tools, and it will be difficult for them to find discover the entire set of tools w/o reading a doc or enabling all tools.

Now that VS Code stable has custom chat mode support (https://code.visualstudio.com/docs/copilot/chat/chat-modes#_custom-chat-modes), what do you think about recommending this instead? (once they fix a couple of bugs where some essential tools don't show up in the picker)

Users can have a mode to support their scenario and quickly switch between modes. Plus, they can use the picker to add additional tools as they are released.

sandeep-sen (approved):

Issue #423, opened 2025-06-19
Support starting the server with multiple services


JasonYeMSFT:
For example, `server start --service <service-a>,<service-b>,<service-c>` to start the server with the 3 services enabled.

PR #424, opened 2025-06-19
Support starting server with multiple services

JasonYeMSFT:
## What does this PR do?

This PR extends the handling logic of --service argument to allow starting the server with multiple services.

## GitHub issue number?

https://github.com/Azure/azure-mcp/issues/423

## Checklist before merging
- [x] **I have read the [contribution guidelines](https://github.com/Azure/azure-mcp/blob/main/CONTRIBUTING.md) on pull request process, code style, and testing.**
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message.  This means that previously merged commits do not appear in the history of the PR.  For more information on cleaning up the commits in your PR,  [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [x] For core features, I have added thorough tests.
- [x] For user-impacting changes (bug fixes, new features, UI/UX changes), I have added a CHANGELOG.md entry linking to this PR.
- [ ] Have a team member run [live tests](https://github.com/Azure/azure-mcp/blob/main/CONTRIBUTING.md#live-tests):
   - [ ] Team Member: Inspect PR for security issues before queueing a test run
   - [ ] Team Member: Add this comment to the pr `/azp run azure - mcp` to start the run

### DIFF
--- a/src/Commands/CommandFactory.cs
+++ b/src/Commands/CommandFactory.cs
@@ -68,7 +68,7 @@ public class CommandFactory
     {
         if (groupNames is null)
         {
-            throw new Exception("groupNames cannot be null.");
+            throw new ArgumentException("groupNames cannot be null.");
         }
         Dictionary<string, IBaseCommand> commandsFromGroups = new();
         foreach (string groupName in groupNames)


CONTRIBUTING.md
```diff
@@ -243,7 +243,7 @@ Get-Process | Where-Object { $_.ProcessName -like "*azmcp*" } | Select-Object Id
 
 To ensure consistent code quality, code format checks will run during all PR and CI builds.
 
-To catch format errors early, run `dotnet format src/AzureMcp.sln` before submitting.
+To catch format errors early, run `dotnet format AzureMcp.sln` before submitting.
```
jongio, line 246: You can just run `dotnet format`

src/Commands/CommandFactory.cs
```diff
@@ -66,17 +66,37 @@ public CommandFactory(IServiceProvider serviceProvider, ILogger<CommandFactory>
 
     public IReadOnlyDictionary<string, IBaseCommand> AllCommands => _commandMap;
 
-    public IReadOnlyDictionary<string, IBaseCommand> GroupCommands(string groupName)
+    public IReadOnlyDictionary<string, IBaseCommand> GroupCommands(string commaSeparatedGroupNames)
     {
-        foreach (CommandGroup group in _rootGroup.SubGroup)
+        var groupNames = commaSeparatedGroupNames.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```
jongio, line 71: I want to make sure that --service isn't required.  It can still be blank and start all right?

CONTRIBUTING.md
```diff
@@ -138,7 +138,7 @@ An optional `--service` parameter can also be set to minimize the number of load
         "server",
         "start",
         "--service",
-        "<service-name>"
+        "<service-name-1>,<service-name-2>"
```
joshfree, line 141: Can you do a grep for --service and update the other references in README, TROUBLESHOOTING, docs/* as appropriate?

CONTRIBUTING.md
JasonYeMSFT: Are you suggesting changing this command to just `dotnet format`?

CONTRIBUTING.md
JasonYeMSFT: I updated one more place in troubleshooting.md to give a hint on the possibility of starting the server with multiple services. Other than that, the existing single service examples are fine.

CONTRIBUTING.md
jongio: Yes, please while you are in there.

src/Commands/CommandFactory.cs
conniey, line 71: Should we be doing the string split here? IIRC, we're using CommandLine options class, which should support this case.

conniey: https://learn.microsoft.com/en-us/dotnet/standard/commandline/syntax#multiple-arguments

wbreza: Agree we should look into this if supported.  Otherwise string split should be `ok` for the short term till this type of arg is supported.

wbreza (approved):
LGTM

src/Commands/CommandFactory.cs
JasonYeMSFT: I can look into implement the multi-service support using the multiple arguments syntax but it will take more time. Are you ok with shipping this as a temporary solution and switch to the new syntax in a follow up PR?

src/Commands/CommandFactory.cs
```diff
@@ -64,17 +64,35 @@ public CommandFactory(IServiceProvider serviceProvider, IEnumerable<IAreaSetup>
...
     public IReadOnlyDictionary<string, IBaseCommand> AllCommands => _commandMap;
 
-    public IReadOnlyDictionary<string, IBaseCommand> GroupCommands(string groupName)
+    public IReadOnlyDictionary<string, IBaseCommand> GroupCommands(string[] groupNames)
     {
-        foreach (CommandGroup group in _rootGroup.SubGroup)
+        if (groupNames is null)
         {
-            if (string.Equals(group.Name, groupName, StringComparison.OrdinalIgnoreCase))
+            throw new Exception("groupNames cannot be null.");
```
conniey, line 71: nit: i believe ArgumentException is better to throw here

wbreza (approved):

JasonYeMSFT:
Could someone help me merge the PR? I don't have the permission to do so.

PR #386, opened 2025-06-17
Enable Telemetry

conniey:
## What does this PR do?

- Adds support for telemetry in azmcp
- Opt-out via AZURE_MCP_COLLECT_TELEMETRY
- Telemetry is enabled if AZURE_MCP_COLLECT_TELEMETRY is not set or set to true.
- `HandleException(CommandContext, ex)` handles setting exceptions in activities.
- `ToolOperations` creates Activity and sets ClientInfo
- `*Command`s that extend from `SubscriptionCommand` set subscription id if possible. 

## Checklist before merging
- [x] **I have read the [contribution guidelines](https://github.com/Azure/azure-mcp/blob/main/CONTRIBUTING.md) on pull request process, code style, and testing.**
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message.  This means that previously merged commits do not appear in the history of the PR.  For more information on cleaning up the commits in your PR,  [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).
- [x] If it's a core feature, I have added thorough tests.
- [x] If it's a noteworthy bug fix or new feature, I have added an entry in CHANGELOG.md with a link back to the PR or issue
- [x] Have a team member run [live tests](https://github.com/Azure/azure-mcp/blob/main/CONTRIBUTING.md#live-tests):
   - [ ] Team Member: Inspect PR for security issues before queueing a test run
   - [ ] Team Member: Add this comment to the pr `/azp run azure - mcp` to start the run

PR #452, opened 2025-06-24
Add VS Code SSO Auth support.

g2vinay:
### Overview

This PR introduces support for authenticating with the Azure account used in the VS Code by reading the `authRecord.json` from the user's `.azure` or `.Azure` profile directory. The new logic is integrated into the `CustomChainedCredential` class.

### Key Changes

- **VS Code Broker Credential Support**:  
  - Added `CreateVsCodeBrokerCredential` to locate and deserialize the VS Code authentication record.
  - Parses and validates the auth record, client ID and tenant ID for safety and correctness.
  - Returns `BrokerCredential` configured for the VS Code context.

- **Credential Chain Update**:  
  - Detects if running under VS Code (`VSCODE_PID`).
  - Prioritizes the VS Code broker credential first in the chain when in VS Code context.
  - By default, VS code broker auth comes DAC and before browser broker credential.

### Motivation

This change enables seamless, single sign-on authentication for users running in VS Code, leveraging their existing Azure sign-in and improving the developer experience.


src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +119,52 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
         return new DefaultAzureCredential(defaultCredentialOptions);
     }
+
+    private static TokenCredential? CreateVsCodeBrokerCredential(string? tenantId)
+    {
+        // VS Code Azure extension client ID
+        const string vsCodeClientId = "aebc6443-996d-45c2-90f0-388ff96faa56";
+        string authRecordPath = Path.Combine(
+            Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+            ".azure",
```
JasonYeMSFT, line 129: The export code allows using either .azure and .Azure directory. Should this code check both kinds of locations?

https://github.com/microsoft/vscode-azureresourcegroups/blob/58822ce9ce5fb1fe4d654a55a6d0eb53117fc656/src/exportAuthRecord.ts#L147

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +119,52 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+            : authRecord?.TenantId;
+
+        var brokerOptions = new InteractiveBrowserCredentialBrokerOptions(0)
+        {
+            ClientId = vsCodeClientId,
+            TenantId = effectiveTenantId,
+            AuthenticationRecord = authRecord,
+            TokenCachePersistenceOptions = new TokenCachePersistenceOptions()
+            {
+                Name = "azure-mcp-msal.cache",
```
JasonYeMSFT, line 161: You can reuse the constant variable `TokenCacheName`

g2vinay: yeah, token cache isn't needed, removed it.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: yeap, that is now added.

JasonYeMSFT (approved):

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -61,9 +61,25 @@ private static TokenCredential CreateCredential(string? tenantId)
             return CreateBrowserCredential(tenantId, authRecord);
         }
 
-        return new ChainedTokenCredential(
-            CreateDefaultCredential(tenantId),
-            CreateBrowserCredential(tenantId, authRecord));
+        var creds = new List<TokenCredential>();
+        var vsCodeCred = CreateVsCodeBrokerCredential(tenantId);
+        bool isVsCodeContext = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSCODE_PID"));
```
joshfree, line 66: This is worth a source code comment on the env variable

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -61,9 +61,25 @@ private static TokenCredential CreateCredential(string? tenantId)
...
+
+        if (isVsCodeContext && vsCodeCred != null)
+        {
+            creds.Add(vsCodeCred);
```
joshfree, line 70: Should we be emitting logging here for future debugging of dogfooder issues?

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +135,76 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
 
         return new DefaultAzureCredential(defaultCredentialOptions);
     }
+
+    private static TokenCredential? CreateVsCodeBrokerCredential(string? tenantId)
+    {
+        const string vsCodeClientId = "aebc6443-996d-45c2-90f0-388ff96faa56";
+        string userProfile = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
```
joshfree, line 142: user profile can be String.Empty on non-Windows operating systems here. Have you tested this?

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +135,76 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+
+        AuthenticationRecord? authRecord;
+        try
+        {
+            using var stream = File.OpenRead(authRecordPath);
+            authRecord = AuthenticationRecord.Deserialize(stream);
+        }
+        catch
+        {
+            // Deserialization failed
```
joshfree, line 162: I feel these cases need to be logged by azmcp for debugging

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +135,76 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+        {
+            ClientId = vsCodeClientId,
+            TenantId = effectiveTenantId,
+            AuthenticationRecord = authRecord
+        };
+
+        return new InteractiveBrowserCredential(options);
+    }
+
+    private static bool IsValidTenantId(string tenantId)
```
joshfree, line 192: This function will incorrectly return true if String.Empty is passed in.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +135,76 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+            if (!IsValidTenantCharacter(c))
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // Helper to validate tenant ID: only hex digits and dashes allowed
+    private static bool IsValidTenantCharacter(char c)
```
joshfree, line 205: Since this is only called by 1 function just move the if-check into the calling method

joshfree (commented):
Left feedback

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +135,76 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+    {
+        return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || (c == '.') || (c == '-');
```
vukelich, line 207: >(c == '.') [](http://example.com/codeflow?start=93&length=10)

What tenant IDs have periods in them?

vukelich, line 207: >c <= 'z' [](http://example.com/codeflow?start=28&length=8)

Only A through F are valid for hex. Why do we check more than that?

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +135,76 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+        // Validate client ID
+        if (!string.Equals(authRecord.ClientId, vsCodeClientId, StringComparison.OrdinalIgnoreCase))
+            return null;
+
+        // Validate tenant ID if present
+        if (!string.IsNullOrEmpty(authRecord.TenantId) && !IsValidTenantId(authRecord.TenantId))
+            return null;
+
+        // Prefer explicit tenantId, else use from auth record
+        string? effectiveTenantId = !string.IsNullOrEmpty(tenantId)
```
vukelich, line 178: >string? effectiveTenantId [](http://example.com/codeflow?start=8&length=25)

Nit: declare this variable as `string` not `string?`. Using `string?` says that the variable could be `null`, but we're doing logic here that ensures the value is not `null`. That means we can use `string` and the compiler will take that into account for finding logical errors.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +135,76 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+            ? tenantId
```
vukelich, line 179: >tenantId [](http://example.com/codeflow?start=14&length=8)

Should we also be validating `tenantId` with the same level of rigor that we do `authRecord.TenantId` by calling `IsValidTenantId`?

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: domain names are allowed to be used as tenant identifiers. 
e.g .contoso.onmicrosoft.com

src/Services/Azure/Authentication/CustomChainedCredential.cs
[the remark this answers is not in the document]
g2vinay: will document it.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: not required, as its validated by the respective credentials that consume it.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: the description comment is incorrect with only hex, removing it.

vukelich:
> public class CustomChainedCredential(string? tenantId = null) : TokenCredential

Since all of the `static` methods will ultimately be called with this member `tenantId` value, it might be worthwhile to have a constructor that validates the value and have member methods that figure out whether to use the VS Code auth file's tenant ID or the one used for this types instantiation.

---
Refers to: src/Services/Azure/Authentication/CustomChainedCredential.cs:21 in 9ac8793. [](commit_id = 9ac8793a560f8ce1f6d73d2303a7ab70b19449c9, deletion_comment = False)

vukelich:
> public class CustomChainedCredential(string? tenantId = null) : TokenCredential

Your call. There's inconsistency on how this `string?` is handled in the various code paths which might be the right design but it'd be good to callout for other maintainers' awareness.

---
In reply to: [3033922386](https://github.com/Azure/azure-mcp/pull/452#issuecomment-3033922386) [](http://example.com/codeflow?ancestors=3033922386)

---
Refers to: src/Services/Azure/Authentication/CustomChainedCredential.cs:21 in 9ac8793. [](commit_id = 9ac8793a560f8ce1f6d73d2303a7ab70b19449c9, deletion_comment = False)

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: I have successfully tested on windows.
As per docs, User Profile returns $HOME on Mac OS and Linux.
https://learn.microsoft.com/en-us/dotnet/maui/macios/system-special-folders?view=net-maui-9.0
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix

I can double check for mac OS

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: updated.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: The tenant ID is validated by the Identity SDK. 
we don't have to do it at this layer.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: tenant ID validation is done in Identity SDK, don't need to do it here.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: tenant ID validation is done in Identity SDK, don't need to do it here.
But for information, tenant IDs can be domain names as well, not just GUIDs.

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: Logging is added.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +140,61 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+        const string vsCodeClientId = "aebc6443-996d-45c2-90f0-388ff96faa56";
+        string userProfile = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+        string authRecordPath = Path.Combine(userProfile, ".azure", "ms-azuretools.vscode-azureresourcegroups", "authRecord.json");
+        if (!File.Exists(authRecordPath))
+        {
+            // Try .Azure if .azure is not present
+            authRecordPath = Path.Combine(userProfile, ".Azure", "ms-azuretools.vscode-azureresourcegroups", "authRecord.json");
+            if (!File.Exists(authRecordPath))
+            {
+                logger?.LogDebug("VS Code Broker Credential -> authRecord.json not found in either .azure or .Azure directory.");
```
vukelich, line 155: >logger?.LogDebug("VS Code Broker Credential -> authRecord.json not found in either .azure or .Azure directory."); [](http://example.com/codeflow?start=16&length=113)

Good add on the logging. Question on it: do we collect it, and if so, how? Or is this just sending data back to the MCP client using the logging messages of the MCP contract?

vukelich: Connie set up OpenTelemetry logging to App Insights in [https://github.com/Azure/azure-mcp/pull/386/files#diff-963bec66e2adf4231f084a3df528681f47b2f52b5e07e26904777f1d868056bf](https://github.com/Azure/azure-mcp/pull/386/files#diff-963bec66e2adf4231f084a3df528681f47b2f52b5e07e26904777f1d868056bf)

On first blush, that means that `ILogger` invocations will be sent to App Insights. If correct, my two thoughts are:

1. We cannot send paths of the user machine to App Insights. That'd be a data handling violation. For a globally central App Insights resource, we can only ever log System Metadata and Organizational Identifiable Information. Other levels in the broad category of "personal data" require much more work. DM me for links to our internal docs on these things.
1. The log messages can be improved to allow better at-scale aggregation by taking advantage of the concept of [structured logging](https://github.com/NLog/NLog/wiki/How-to-use-structured-logging) which is a big feature for [ILogger in .NET](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line). App Insights, when exporting `ILogger` messages, will take log format argument and add a child in the `Properties` column in the resulting App Insights row for the log along with the message (depends on the version of the Azure Monitor SDK if the final formatted message or original message).

For the GitHub Copilot for Azure VS Code extension, we made a similar complication in design choice that I'd strongly advise against repeating. That design is to use `ILogger` for both "logging we want the client to see and upload" and "logging we want to send to our automatic telemetry collection". GHCPAZ handled this by effectively allow-listing the OTel `LogRecord`s could go to the Microsoft-internal logging for VS Code and extensions. It was an opt-in mechanism per invocation of `ILogger` that determined if it could go to both places or only to the Output window in VS Code. This could be accomplished because we were creating our own OTel exporter as a wrapper around an internal logging lobrary rather than using App Insights' direct integration with OTel.

vukelich: I chatted with Connie, and she's confirmed that `ILogger` events are going to our App Insights resource. She's going to look into wholly disable any `ILogger` messages going through the App Insights OpenTelemetry integration in the short-term for this usage of `ILogger` to be data handling compliant. Josh, a more thorough set of planning will be needed to reintroduce single-message logging for our App Insights resource. 

We'll continue to transmit all data attached to [.NET Activity objects](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-instrumentation-walkthroughs#activity) as part of our App Insights + OTel integration. In explicit terms, I mean we'll continue to send data to App Insights for anything created and added to via our custom [ITelemetryService](https://github.com/Azure/azure-mcp/blob/9f8b411e897820124b9b21b1ae02715dcfa4bc15/src/Services/Telemetry/ITelemetryService.cs#L9).

joshfree: I'd wait on the ILogger change from @conniey and then merge this after

joshfree: Related: https://github.com/Azure/azure-mcp-pr/issues/294

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: added the defensive check.

src/Services/Azure/Authentication/CustomChainedCredential.cs
joshfree: Update: https://github.com/Azure/azure-mcp/pull/540 was merged yesterday.  Marking this conversation Resolved to unblock the PR.

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +140,66 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+        if (authRecord is null)
+        {
+            logger?.LogDebug("VS Code Broker Credential -> Deserialized VS Code AuthenticationRecord is null.");
+            return null;
+        }
+
+        // Validate client ID
+        if (!string.Equals(authRecord.ClientId, vsCodeClientId, StringComparison.OrdinalIgnoreCase))
+        {
+            logger?.LogDebug("VS Code Broker Credential -> VS Code AuthenticationRecord clientId mismatch. Expected {Expected}", vsCodeClientId);
```
joshfree, line 186: Try running dotnet format, these lines might be too long.

joshfree (approved):

src/Services/Azure/Authentication/CustomChainedCredential.cs
g2vinay: dotnet format auto runs on git push

.vscode/cspell.json
```diff
@@ -259,6 +259,8 @@
         "WINDOWSPOOL",
         "WINDOWSVMIMAGE",
         "winget",
-        "Xunit"
+        "Xunit",
+        "azuretools",
```
vukelich, line 263: Looks like other entries are sorted alphabetically in this list. Do we want to continue that?

src/Services/Azure/Authentication/CustomChainedCredential.cs
```diff
@@ -119,4 +140,66 @@ private static DefaultAzureCredential CreateDefaultCredential(string? tenantId)
...
+
+        var options = new InteractiveBrowserCredentialBrokerOptions(0)
+        {
+            ClientId = vsCodeClientId,
+            TenantId = effectiveTenantId,
+            AuthenticationRecord = authRecord
+        };
+
+        logger?.LogDebug("VS Code Broker Credential -> Successfully created VS Code Broker Credential");
+        return new InteractiveBrowserCredential(options);
```
vukelich, line 203: Out of curiosity, will this credential create an access token with MFA? It might be important for the upcoming change to [ARM](https://learn.microsoft.com/en-us/entra/identity/authentication/concept-mandatory-multifactor-authentication?tabs=dotnet). I'm still trying to fully wrap my head around the implications for our products.

src/Services/Azure/Subscription/SubscriptionService.cs
```diff
@@ -1,15 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+
```
vukelich, line 4: Nit: line can be removed

vukelich (approved):
Minor comments. Skippable.

src/Services/Azure/Authentication/CustomChainedCredential.cs
xiangyan99: > As I understand things
> 
> 1. A user logs in to VS Code using their username/password or WAM
-> Tyler told me VSCode would only use WAM if it is running on windows.
> 2. They receive a non-MFA token
> 3. Each time the agent attempts a create, update, or delete operation on Azure resources, a step-up challenge (e.g., re-sign in / MFA) will be required, causing the script/agent to be interrupted and the existing cached token to be invalidated
-> It is required only for the first challenge. The MFA token will be cached, eliminating the need for further re-auth.

some comments inline. (assuming in the scope of Azure MCP server and az-cli-extension is not used)

xiangyan99: > Out of curiosity, will this credential create an access token with MFA? It might be important for the upcoming change to [ARM](https://learn.microsoft.com/en-us/entra/identity/authentication/concept-mandatory-multifactor-authentication?tabs=dotnet). I'm still trying to fully wrap my head around the implications for our products.

This request will obtain a token using the default settings.

If MFA is enforced at the tenant level (e.g., MSFT), an MFA token will be retrieved; otherwise, a non-MFA token is obtained.

I am collaborating with the Entra team to explore options that would allow users to proactively opt in to receive an MFA token.

joshfree: Yup, only on Windows until we ship NodeJS WAM BrokeredAuth packages for macOS and Linux. This scenario here is how we got MSAL to fund the Linux package for later this year (Minh-Anh's feature)

vukelich: > If MFA is enforced at the tenant level (e.g., MSFT), an MFA token will be retrieved; otherwise, a non-MFA token is obtained.

OK, this is good to know. @xiangyan99 , continuing on our conversation from Office Hours this week, I wonder if there's a way to configure desktop client sign-ins for first party apps to require MFA sign-ins. This would mean we never need to worry about the challenge loop. I can only find Conditional Access policies as an option for third party apps, and I'm looking through the FPA docs to see if there's anything special there. What do you think?

conniey (approved):

PR #475, opened 2025-06-26
Fix AOT Analysis by Adding DisableWarningsAsErrors Property

anuchandy:
AOT (Ahead-of-Time) compatibility analysis was failing after enabling `TreatWarningsAsErrors` in the build configuration in this [pr](https://github.com/Azure/azure-mcp/pull/435). Since AOT analysis intentionally generates IL warnings to identify potential compatibility issues, these warnings were being treated as errors, causing the analysis to fail prematurely.

To address this, `Analyze-AOT-Compact.ps1` will pass this property `TreatWarningsAsErrors ` as `false` to the dotnet publish command which runs the AOT analysis.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,6 +3,8 @@
     <TargetFramework>net9.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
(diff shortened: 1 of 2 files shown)


JonathanCrd (approved):
LGTM :)

PR #904, opened 2025-08-08
Refactor storage account handling to return metadata

### LINKED ISSUE #899: Update `storage account list` to return more storage account metadata.
Right now it just returns the storage account name.  Let's get it updated to return more info about the storage account.

jongio:
Update the `GetStorageAccounts` method to return detailed metadata objects instead of strings, enhancing the `AccountListCommand` and improving related tests and documentation. This change ensures better data representation and usability for storage account information.

Fixes #899


.vscode/tasks.json
```diff
@@ -1,17 +1,33 @@
...
+			"type": "process",
+			"args": [
+				"build",
+				"${workspaceFolder}/AzureMcp.sln",
+				"/property:GenerateFullPaths=true",
+				"/consoleloggerparameters:NoSummary"
+			],
+			"problemMatcher": "$msCompile"
+		},
+		{
```
jongio, line 16: Copilot added this block because the type dotnet wasn't working on my machine.  This look identical to the one above with a diff type.

alzimmermsft (approved):

Issue #756, opened 2025-07-28
Feature: Integrate ARG management library to retrieve all ARM resources


ArthurMa1978:
## Proposal

This is a proposal to reduce the size of the AzureMCP server.
The `Azure.ResourceManager.ResourceGraph` library enables retrieval of any ARM resource data, making it a viable replacement for specific `Azure.ResourceManager.XXX` libraries used in individual tools.

## Details

**Current Approach**
Each tool currently relies on its corresponding .NET management library to list or read Azure resource data. For example, the List AKS Cluster tool uses the `Azure.ResourceManager.ContainerService` library. If we introduce a new tool, such as one to list MySQL Flexible Servers, we would need to add a dependency on `Azure.ResourceManager.MySql`.

As more tools are added, the number of Azure management .NET SDKs integrated into the project will grow, leading to a continuous increase in the size of the AzureMCP server.

**Proposed Optimization**
The `Azure.ResourceManager.ResourceGraph` (ARG) library supports querying any ARM resource. It can be used to replace the read operations currently handled by individual service-specific libraries. For instance, the List AKS Cluster tool can be refactored to use a Resource Graph query instead of relying on the ContainerService SDK:

``` C#
"Resources | where type =~ 'Microsoft.ContainerService/managedClusters' | project id, name, type, location, tags, sku, properties"
```

## Pros & Cons

✅ **Pros**

**Reduced Server Size:** Significantly decreases the overall size of the AzureMCP server by minimizing dependencies.
**Improved Query Performance:** Resource Graph queries are generally faster than service-specific SDK calls.

⚠️ **Cons**

**Latency in Resource Updates:** ARG may not reflect real-time changes immediately, leading to potential delays in resource state visibility.
**Manual Deserialization Required:** No strongly typed models, results must be manually parsed and deserialized.

joshfree:
Issue has been migrated to the https://github.com/microsoft/mcp repo.

PR #938, opened 2025-08-12
Using ARG query to support Sql resource read operations

ArthurMa1978:
## What does this PR do?

This PR addresses part of the issue tracked in https://github.com/Azure/azure-mcp/issues/756 by updating the SQL tool to eliminate its dependency on Azure.ResourceManager.Sql.
As a result, it reduces the AzureMCP agent size by approximately 6.41 MB.

## GitHub issue number?

https://github.com/Azure/azure-mcp/issues/756

## Pre-merge Checklist

- [x] Required for All PRs
    - [ ] **Read [contribution guidelines](https://github.com/Azure/azure-mcp/blob/main/CONTRIBUTING.md)**
    - [ ] PR title clearly describes the change
    - [ ] Commit history is clean with descriptive messages ([cleanup guide](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md))
    - [ ] Added comprehensive tests for new/modified functionality
    - [ ] Updated `CHANGELOG.md` for product changes (`features, bug fixes, UI/UX, updated dependencies`)
    - [ ] Spelling check passes: `.\eng\common\spelling\Invoke-Cspell.ps1`
- [ ] For MCP tool changes:
    - [ ] Updated `README.md` documentation
    - [ ] Updated command list in `/docs/azmcp-commands.md`
    - [ ] Updated test prompts in `/e2eTests/e2eTestPrompts.md`
    - [ ] For new or modified tool descriptions, ran the `eng/tools/ToolDescriptionConfidenceScore` tool and obtained a result >= 0.4
- [ ] 👉 For Community (non-Azure team member) PRs:
    - [ ] **Security review**: Reviewed code for security vulnerabilities, malicious code, or suspicious activities before running tests (`crypto mining, spam, data exfiltration, etc.`)
    - [ ] **Manual tests run**: added comment `/azp run azure - mcp` to run *Live Test Pipeline*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,6 @@ The Azure MCP Server updates automatically by default whenever a new release com
 
 ### Features Added
 
-
 - Added support for the following Azure Deploy and Azure Quota operations: [[#626](https://github.com/Azure/azure-mcp/pull/626)]
   - `azmcp_deploy_app_logs_get` - Get logs from Azure applications deployed using azd.
   - `azmcp_deploy_iac_rules_get` - Get Infrastructure as Code rules.


jongio:
@copilot review this PR against checklist and guidance from new-command.md

areas/sql/src/AzureMcp.Sql/Services/SqlService.cs
jongio, line 1: Consolidated PR Review (Guided by docs/new-command.md checklist)

Overall: Good refactor replacing direct Azure.ResourceManager.Sql usage with Resource Graph to reduce footprint. BaseAzureResourceService abstraction is a nice reuse candidate. A few consistency, safety, naming, and doc/test gaps need attention before merge.

Checklist Mapping:
1. Package cleanup: ✅ Removed Azure.ResourceManager.Sql from Directory.Packages.props and project. Added Azure.ResourceManager.ResourceGraph in core. Consider removing if already centrally declared (verify duplication). Ensure no lingering transitive-only dependency assumptions.
2. Service base class choice: ✅ SqlService now inherits BaseAzureResourceService. Need to confirm no remaining write operations rely on the removed SDK (only read ops appear here).
3. AOT safety: ⚠ Internal model classes added (SqlDatabaseData, etc.) should be included in source gen context. You added them to SqlJsonContext — good. Confirm no reflection-based deser paths outside source gen; FromJson uses context ✔.
4. Naming / internal models: Consider suffix “Data” consistency (all have ...Data) – OK. But DatabaseIdentity lacks JsonSerializable attribute in context (not needed if only nested) – OK.
5. Escaping / KQL injection: Added EscapeKqlString in BaseAzureService and applied in Workbooks changes and some Sql queries. However:
   - In BaseAzureResourceService.ExecuteSingleResourceQueryAsync the queryFilter does NOT escape resourceType/resourceGroup (unlike ExecuteResourceQueryAsync) and does not use EscapeKqlString at all. Use same escaping for parity.
   - Additional filters passed in (e.g. name =~ '{EscapeKqlString(databaseName)}') are escaped correctly in caller but ExecuteSingleResourceQueryAsync’s own filter pieces should still call EscapeKqlString.
6. Consistency: In ExecuteSingleResourceQueryAsync you build: "Resources | where type =~ '{resourceType}' and resourceGroup =~ '{resourceGroup}'" vs ExecuteResourceQueryAsync uses EscapeKqlString. Align.
7. Null-safety: Converter methods assume Properties not null. Defensive null-checks or throw w/ descriptive message would be safer (Properties could be missing in partial graphs). E.g. sqlDatabase.Properties?.Status.
8. Performance: Each query calls tenantService.GetTenants() then FirstOrDefault per request. Consider caching tenant lookup per subscription in BaseAzureResourceService (dictionary) to avoid repeated full tenant enumeration.
9. Error handling: SqlService catch blocks just log and rethrow. This bubbles raw exceptions; existing pattern typically transforms into standardized responses higher up, but consider mapping KeyNotFoundException to 404 in command layer or override GetErrorMessage/Status if commands rely on service exceptions. Ensure commands already handle (not shown in diff).
10. Elastic Pool mapping: Edition currently derived from elasticPool.Sku; but Edition property is not set on SqlElasticPool (only licenseType etc.). Confirm parity with old model (previous code didn’t set Edition either—OK).
11. Removed fields: DatabaseDtuMin/Max, Dtu, StorageMB intentionally null with comment. Consider documenting in docs/azmcp-commands.md that these aren’t available via ARG.
12. Live tests: Adjusted only type comparisons to ignoreCase; ensure tests still cover negative scenarios and error handling (no changes shown for 404 / missing resource). Consider adding a test for non-existent database returning 404 scenario after refactor.
13. Documentation: Updated new-command.md with Resource Graph guidance – ✔. Need to also update CHANGELOG.md (not in diff), README.md, docs/azmcp-commands.md, e2eTests/e2eTestPrompts.md if tool descriptions or output semantics changed (e.g., potential casing differences or omitted properties). Checklist in PR body not yet checked.
14. Spelling list: Added "cutover" and "Hyperscale" – ✔. Confirm no other new terms flagged by cspell (run script).
15. Source generation context: Added SqlDatabaseData, SqlServerAadAdministratorData, SqlElasticPoolData, SqlFirewallRuleData but NOT SqlDatabaseProperties, SqlElasticPoolProperties, etc. Because they’re nested, that’s fine if only deserialized via parent context. Verify no direct serialization of nested property types alone (e.g., unit tests expecting typed result) — if needed, add explicit annotations.
16. Access Modifiers: All new model classes are internal sealed – ✔.
17. Base class utility: EscapeKqlString returns empty string on null/empty input; consider preserving null vs empty semantics (currently you treat null param as empty filter part—OK). Document expected behavior.
18. Potential bug: In ExecuteResourceQueryAsync you build query using resourceType escaped; In ExecuteSingleResourceQueryAsync you didn’t escape. Also additionalFilter uses string interpolation directly; caller should guarantee safe content, but maybe document requirement or perform limited validation (reject unbalanced quotes) to reduce risk.
19. Logging: When throwing KeyNotFoundException, maybe log at Information or Warning first for diagnostics with identifiers.
20. Cancellation tokens: Ensure that ExecuteResourceQueryAsync / ExecuteSingleResourceQueryAsync propagate cancellationToken to GetSubscription (currently not), only to GetResourcesAsync. Pass cancellationToken into GetSubscription to be fully responsive.
21. Tenants enumeration cost: If multi-subscription/tenant heavy use, consider a dedicated IAzureGraphClient abstraction to reuse tokens and skip repeated GetTenants calls.
22. Unit tests: None shown in diff for new base class. Suggest adding dedicated tests for BaseAzureResourceService (mocking Graph response) or at least augment SqlService unit tests to validate query building (with special chars in names to validate escaping). At minimum, add a test with serverName containing a single quote to ensure no KQL breakage.
23. AOT Trimming: Verify no reflection by System.Text.Json outside source-gen types. FromJson uses context; good. Ensure no raw JsonSerializer.Deserialize<T>(jsonString) of generic T without context in new code.
24. Style: Minor nit - trailing comments like // Read the JSON response content and create a model instance from it. appear multiple times; could be DRY, but optional.
25. DRY: Converter methods could be local functions or static lambdas reused; acceptable as-is.

Actionable Recommendations:
- [ ] Escape resourceType/resourceGroup in ExecuteSingleResourceQueryAsync via EscapeKqlString.
- [ ] Use EscapeKqlString consistently inside ExecuteResourceQueryAsync (already done) and ensure additionalFilter isn’t null-concatted without space (currently using " and {additionalFilter}" OK).
- [ ] Add null-safe navigation for Properties in converter methods (or explicit throw with message) to avoid NRE if ARG returns partial data.
- [ ] Pass cancellationToken into subscriptionService.GetSubscription in both query helpers.
- [ ] Consider caching tenantResource lookup.
- [ ] Add/update docs (CHANGELOG.md, azmcp-commands.md) to note ARG-based implementation & potential missing DTU fields.
- [ ] Add negative path live test for missing database (expect 404).
- [ ] Add unit test for KQL escaping with single quote in names.
- [ ] Confirm no remaining direct ARM operations require removed package.
- [ ] (Optional) Add explicit JsonSerializable attributes for nested property classes if future direct serialization is anticipated.

Nice work on reducing package size and centralizing query logic. Addressing the listed items will align fully with new-command.md guidance and strengthen robustness.

jongio (commented):
Added consolidated review feedback referencing new-command.md checklist.

areas/sql/src/AzureMcp.Sql/Services/SqlService.cs
```diff
@@ -26,44 +26,21 @@ public class SqlService(ISubscriptionService subscriptionService, ITenantService
...
-            var sqlServerResource = await resourceGroupResource.Value
-                .GetSqlServers()
-                .GetAsync(serverName);
-
-            var databaseResource = await sqlServerResource.Value
-                .GetSqlDatabases()
-                .GetAsync(databaseName);
-
-            var database = databaseResource.Value.Data;
+            var result = await ExecuteSingleResourceQueryAsync(
```
JasonYeMSFT, line 29: If I remember correctly, ARG query has a hard limit on how many items it can return in one result. The old implementation uses GetAllAsync() which I suppose internally enumerates all the items before returning. I don't see the BaseAzureResourceService class handling paging for enumerating paged ARG results. Should we document this as a breaking change?

.vscode/cspell.json
```diff
@@ -86,6 +86,7 @@
                 "cslschema",
                 "codeql",
                 "codesign",
+                "cutover",
```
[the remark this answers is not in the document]
ArthurMa1978: Changelog updated

areas/sql/src/AzureMcp.Sql/Services/SqlService.cs
ArthurMa1978: Good point. After discussing with Xiang and following Jeffery's recommendation, I’ve set a default limit of 50 for the ARG query. I’ll update this method to support pagination once a conclusion is reached in https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/799.

areas/sql/src/AzureMcp.Sql/Services/SqlService.cs
ArthurMa1978: Great! I've made updates based on the review feedback.

areas/sql/src/AzureMcp.Sql/Services/SqlService.cs
jongio: Let's not hold on 799 for pagination.  I don't mind doing 50 for now, but the gate isn't 799, we can design it and implement it as we see fit - independent of what is in the mcp protocol.

CHANGELOG.md
```diff
@@ -6,12 +6,24 @@ The Azure MCP Server updates automatically by default whenever a new release com
 
 ### Features Added
 
+- Introduced `BaseAzureResourceService` class that uses Azure Resource Graph queries for improved performance and reduced API calls when retrieving Azure SQL resources. This provides faster and more efficient resource discovery compared to direct ARM API calls.
```
jongio, line 9: You need to link to the PR.  See other changelog entries for example

CHANGELOG.md
```diff
@@ -6,12 +6,24 @@ The Azure MCP Server updates automatically by default whenever a new release com
...
+- Fixed SQL service test assertions to use case-insensitive string comparisons for resource type validation.
+- Fixed HttpClient service test assertions to properly validate NoProxy collection handling instead of expecting a single string value.
+
 ### Other Changes
 
+- Refactored SQL service implementation to use Azure Resource Graph queries instead of direct ARM API calls, significantly improving performance and reducing the number of API requests required for SQL resource operations.
+- Removed dependency on `Azure.ResourceManager.Sql` package by migrating to Azure Resource Graph queries, reducing package size and improving startup performance.
+- Added `BaseAzureResourceService` abstract class that provides common Resource Graph query functionality with tenant caching and consistent error handling for Azure services.
+- Enhanced `BaseAzureService` with `EscapeKqlString` method for safe KQL query construction across all Azure services.
+- Updated development documentation to include guidance on choosing between `BaseAzureResourceService` and `BaseAzureService` based on service requirements.
```
jongio, line 25: Let's reduce the text in changelog.  You can put under a main bullet with a couple of points.  This is too verbose for a changelog.

areas/sql/src/AzureMcp.Sql/Commands/Database/DatabaseShowCommand.cs
```diff
@@ -53,13 +53,6 @@ public override async Task<CommandResponse> ExecuteAsync(CommandContext context,
                 options.Subscription!,
                 options.RetryPolicy);
 
-            if (database == null)
```
jongio, line 56: Why removed?

areas/sql/src/AzureMcp.Sql/Services/Models/SqlServerAadAdministratorData.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// <auto-generated/>
+
+#nullable enable
```
jongio, line 6: Please comb for these

core/src/AzureMcp.Core/Services/Azure/BaseAzureResourceService.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// </summary>
+    private readonly ConcurrentDictionary<Guid, TenantResource> _tenantCache = new();
+
+    /// <summary>
+    /// Gets the tenant resource for the specified subscription, using caching to avoid repeated tenant enumeration.
+    /// </summary>
+    /// <param name="tenantId">The tenant ID from the subscription</param>
+    /// <param name="cancellationToken">Cancellation token</param>
+    /// <returns>The tenant resource associated with the subscription</returns>
+    private async Task<TenantResource> GetTenantResourceAsync(Guid? tenantId, CancellationToken cancellationToken = default)
```
jongio, line 39: TenantService should handle tenant caching.  See SubscriptionService for example

eng/vscode/README.md
```diff
@@ -88,8 +88,11 @@ Here are some cool prompts you can try across our supported Azure services:
 * "List containers in my Cosmos DB database"
```
jongio, line 88: The diff on this is weird.  Make sure it is formatted correctly

jongio (changes requested):

CHANGELOG.md
ArthurMa1978: updated

areas/sql/src/AzureMcp.Sql/Services/Models/SqlServerAadAdministratorData.cs
ArthurMa1978: Removed

areas/sql/src/AzureMcp.Sql/Commands/Database/DatabaseShowCommand.cs
ArthurMa1978: The `GetDatabaseAsync` does not return null when the database is missing; instead, it throws a `KeyNotFoundException`. As a result, the `GetErrorMessage` and `GetStatusCode` methods are updated to handle this scenario as a 404 error.

core/src/AzureMcp.Core/Services/Azure/BaseAzureResourceService.cs
ArthurMa1978: I just realized the GetTenants already has caching, so remove the cache here

eng/vscode/README.md
ArthurMa1978: yes, it's weird, there is no format change, maybe this is a github comparation defect

ArthurMa1978: Reverted this and deferred for future action

areas/sql/src/AzureMcp.Sql/Services/Models/SqlDatabaseProperties.cs
```diff
@@ -0,0 +1,153 @@
...
+
+using System.Text.Json.Serialization;
+
+namespace AzureMcp.Sql.Services.Models
+{
+    /// <summary>
+    /// A class representing the SqlDatabase properties model.
+    /// A database resource properties.
+    /// </summary>
+    internal sealed class SqlDatabaseProperties
```
jongio, line 12: I'm thinking that we may want to remove any properties that we think the user will never query on or be interested in.  Is that possible to determine with some confidence?  (Looking to trim code and save on binary size as a general rule)

areas/sql/src/AzureMcp.Sql/Services/Models/SqlFirewallRuleProperties.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// A class representing the SqlFirewallRule properties data model.
+    /// A server firewall rule properties.
+    /// </summary>
+    internal sealed class SqlFirewallRuleProperties
+    {
+        /// <summary> The start IP address of the firewall rule. Must be IPv4 format. Use value '0.0.0.0' for all Azure-internal IP addresses. </summary>
+        [JsonPropertyName("startIpAddress")]
+        public string? StartIPAddress { get; set; }
+        /// <summary> The end IP address of the firewall rule. Must be IPv4 format. Must be greater than or equal to startIpAddress. Use value '0.0.0.0' for all Azure-internal IP addresses. </summary>
+        [JsonPropertyName("endIpAddress")]
```
jongio, line 18: do we need these attributes

jongio (approved):

PR #971, opened 2025-08-14
Add MCP server config options to VSIX

g2vinay:


eng/vscode/package.json
```diff
@@ -53,6 +53,17 @@
                     "type": "array",
                     "items": { "type": "string" },
                     "description": "List of Azure MCP services to enable. If empty or unset, all services are enabled by default."
+                },
+                "azureMcp.mode": {
+                    "type": "string",
+                    "enum": ["namespace", "all", "single"],
+                    "default": "namespace",
+                    "markdownDescription": "MCP server exposure mode. 'namespace' collapses tools per service (default), 'all' exposes every tool individually, 'single' exposes one 'azure' tool that routes internally."
```
joshfree, line 61: Reorder this to be

"single", "namespace (default)", "all"

MCP Tool exposure mode.  `single` collapses every tool (100+) into one (1) single tool that routes internally., `namespace` collapses all tools down into logical Azure service namespace grouping (about 30 tools), while `all` exposes every mcp tool directly to the MCP Client.  We recommend `namespace` as the right balance between MCP tool count and tool selection accuracy."

joshfree (changes requested):

eng/vscode/package.json
g2vinay: updated

eng/vscode/package.json
```diff
@@ -53,6 +53,17 @@
                     "type": "array",
                     "items": { "type": "string" },
                     "description": "List of Azure MCP services to enable. If empty or unset, all services are enabled by default."
+                },
```
joshfree, line 56: Is it possible to change how these items are displayed?

Image

JasonYeMSFT: Last time I looked at it the answer was no. VS Code infers the title from the configuration id and there is no way to customize them.

g2vinay: yeah, we're keeping the prefix, as it is also helpful when working with settings in raw .json

JasonYeMSFT (approved):

joshfree (approved):
Thank you.  The dropdown and text changes look good

PR #992, opened 2025-08-15
Move e2eTestPrompts.md to /docs

jongio:
This was hanging out in the root.  I'd like to see it moved to docs/


vcolin7:
> Do we also need to update Victor's tool to reference the new location? @vcolin7

Yes, we'll have to update the tool code to look for the new prompts file.

xiangyan99 (approved):

PR #995, opened 2025-08-15
Add Azure MCP Server Hosting Specification document

jongio:
Introduce a new document outlining the specifications for hosting Azure MCP servers.

### DIFF
--- a/docs/Hosting.md
+++ b/docs/Hosting.md
@@ -9,40 +9,225 @@
 | Field | Value |
 |-------|-------|
 | Title | Azure MCP Server Hosting Specification |
-| Version | 0.2-draft |
+| Version | 0.5-draft |
 | Status | Draft (for internal review) |
 | Owners | TBD |
-| Last Updated | 2025-08-15 |
+| Last Updated | 2025-08-19 (Microsoft-hosted model integrated; Sections 24/25 promoted to normative Phase 0 for hosted) |
 | Target Milestone | Remote Preview |
 
 All normative requirements use RFC 2119 keywords (MUST, SHOULD, MAY, MUST NOT, SHOULD NOT).
 
+### Executive Summary
+**One unified blueprint** to run the Azure MCP Server in Azure: immediately safe for customer self‑hosting and deliberately low risk when Microsoft hosts it. The design ties every real Azure action to an authenticated user, constrains privilege to only what each tool declares, and adds capability in controlled phases. We start with a strictly static, non‑tenant‑aware “safe” tool subset (no live calls into customer Azure, no secrets, no subscription or resource IDs), add delegated read access only after security gates pass, then introduce isolated write tiers. **No hidden elevation paths. No speculative features before controls.**
+
+
+**Core Proposal**
+| Aspect | Summary | Value | Guardrails |
+|--------|---------|-------|------------|
+| Self‑Hosted Baseline | All real Azure calls use per‑user delegated tokens (Auth Code + PKCE + OBO) from day one | Immediate productive use with precise RBAC & audit fidelity | MI restricted to infra support (secrets, logging) only; no privilege aggregation |
+| Hosted Phase 0 | Public endpoint offers only static, pre‑vetted safe tools; zero interaction with customer Azure resources | Enables early ecosystem integration with negligible customer risk | Prominent labeling, telemetry, strict allowlist, hard rate limits |
+| Phase Progression | Read (Phase 1) then constrained write (Phase 2) added only after explicit security & reliability gates | Predictable risk escalation; stakeholders see measurable readiness before capability expands | Exit metrics + rollback objective (<15 min) + mandatory scope minimization tests |
+| Unified Specification | Single doc covers self‑host + hosted; conditional exposure instead of forked designs | Eliminates drift, duplicated reviews, inconsistent guidance | Any change merges once; phase gating controls runtime activation, not content divergence |
+
+**Problem Statement**: Existing ad hoc hosting paths trade off velocity and security (broad Managed Identity roles, privilege expansion, speculative features). This design replaces that with per‑request least‑privilege delegation, explicit phase gates, and a static safe surface for early hosted discovery.
+
+**Strategic Guarantees**
+1. No Azure resource access occurs without an authenticated user principal (except static safe subset).  
+2. Every operation’s required scopes are declared up front and validated; over‑broad requests are rejected.  
+3. Advancing a hosted phase never removes an existing control (monotonic security).  
+4. Rollback from any new capability to the prior phase target is operational in <15 minutes (artifact + config only; no data migration).  
+5. MI role set remains constant across phases (infra-only), preventing silent privilege creep.
+
+**Rationale**: Enables immediate self‑hosted use while establishing measurable gates for hosted expansion, reducing tension between rapid delivery and required controls.
+
+**Key ideas 🧩**
+* ⚙️ Stateless Implementation: Although the MCP protocol defines conversational / session semantics, our server treats each request independently (no server-held workflow/session state) so any replica can handle any request.
+* 🎯 Declarative scopes: Each tool declares only the minimal Azure scopes it needs (or marks itself static); broader requests are rejected.
+* 🛑 Fail‑fast startup: Missing required auth configuration halts the server (unless explicitly in safe mode).
+* 🧱 Tiered model: Safe (static), Guarded (read), Privileged (write). Only Safe runs without auth.
+* 🔌 Modular infra: Add logging, tracing, network isolation, cache, Key Vault only when needed.
+
+**Tracks 🗺️**
+1. Self‑Hosted (single tenant): full per‑user delegated calls now.
+2. Hosted: Phase 0 safe tools only; later phases add delegated access, isolation, and privileged operations.
+
+**Security at a glance 🔐**
+* ❌ No MI shortcut for user actions.
+* 🔄 Per‑user tokens validated then exchanged (OBO) for exact downstream scopes.
+* 📜 Rate limiting + logging from day zero; PoP / stronger token binding evaluated later.
+* 🪧 Safe mode clearly labeled to prevent mistaken Azure interaction assumptions.
+
+**Success indicators 📊**
+* ✅ <1% unauthorized attempts succeed.
+* 🚀 P95 latency <2s (delegated) / <1.2s (safe tools).
+* 🔁 Risky feature rollback <15 minutes.
+
+**Headline risks & mitigations ⚠️**
+| Risk | Simple Mitigation |
+|------|------------------|
+| Shared public client misuse | Monitor telemetry; encourage vendor-specific client IDs |
+| Token replay | Enforce TLS; short lifetimes; plan for PoP later |
+| Managed Identity overreach | MI barred from user authorization paths |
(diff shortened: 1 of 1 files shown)


joshfree (approved):

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+2. Full parity across all Azure compute offerings.
+3. Advanced stateful collaboration or session migration.
+4. Deep non-Azure cloud hardening guidance.
+5. Immediate support for dynamic client registration (blocked by current Microsoft Entra platform capabilities).
+
+## 3. Definitions
+| Term | Definition |
+|------|------------|
+| MCP | Model Context Protocol. |
+| PRM Document | Initial protocol registration metadata from the server. |
```
localden, line 45:
```suggestion
| PRM Document | Protected Resource Metadata ([RFC 9728](https://datatracker.ietf.org/doc/rfc9728/)). |
```

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+6. SHOULD periodically (e.g., every 15 minutes) log a security heartbeat summarizing the same (to aid forensic review).
+
+### 6.5 Managed Identity Scope Strategy
+* Principle of least privilege: Create MI at resource group scope where possible.
+* Prefer data-plane read roles first (e.g., Storage Blob Data Reader) vs full contributor roles.
+* Document escalation path if write capabilities required.
+
+### 6.6 Future: Per-user Delegation
+* Evaluate OAuth 2.0 OBO with confidential app (App Registration + audience validation).
+* Consider Proof-of-Possession (TLS-bound tokens) if MCP spec adds channel binding.
```
localden, line 118: What is this in relation to (which part of the flow)?

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+5. Client registers with the Authorization Server (dynamic client registration) OR already possesses a suitable client ID.
+6. Client initiates Authorization Code Flow with PKCE directly against Entra (user authenticates + consents).
+7. Entra issues an access token whose `aud` (audience) is the MCP server (plus optional ID token for client state).
+8. Client sends that access token to the MCP server on each request (e.g., Bearer token header / agreed channel field).
+9. MCP server validates token (issuer, signature, expiry, audience, scopes / roles, nonce / c_hash as needed).
+10. MCP server performs OAuth 2.0 On-Behalf-Of (OBO) flow to exchange the user token for downstream resource tokens (ARM, Storage, etc.).
+11. MCP server maps/associates downstream tokens to the inbound user principal (cache keyed by user object ID + scopes) and executes tool operations.
+12. Responses returned; auditing & telemetry capture per-user context.
+
+**Current Blocking Gap**: No dynamic client registration for arbitrary third-party native MCP clients.
```
localden, line 175: This is actually going to be your biggest problem. Because Entra ID doesn't support DCR, you will need to either (1) implement the AS part yourself for this, which puts you on the path of a token factory, or (2) ask every user to enter a client ID, which is, unfortunately, a pretty bad experience usability-wise.

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+
+**Interim Alternatives**
+| Approach | Description | Pros | Cons / Risks |
+|----------|-------------|------|--------------|
+| Pre-registered multi-tenant Public Client | Publish a client ID (no secret) that all MCP clients can use for auth code + PKCE | Simple for ecosystem, no secret distribution | Hard to apply client-specific policies, revocation affects all, potential misuse rate limits |
```
localden, line 180: I don't think you can get away with this, unfortunately. What you're saying here is that you will maintain an app registration for every single MCP client that are not Microsoft, and the security team will shut that down fairly quickly. For this to work, you'd need to set up a client with multiple redirect URIs, etc. that are completely out of your control.

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+| Per-vendor Client IDs (manual registration) | Each MCP client vendor maintains an Entra app registration | Stronger isolation/accountability | Friction for new clients; coordination overhead |
```
localden, line 181: I mean, ideally yes - but good luck convincing every vendor (e.g., Anthropic, Cursor) to go an create an Azure tenant and register their Entra ID app.

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+| Device Code Flow (interactive fallback) | Clients invoke device code to let user authenticate in browser | Works without embedded browser, simpler UX for CLIs | Less seamless, slower, not ideal for high-frequency re-auth scenarios |
```
localden, line 182: But you'd still need a client registration for this?

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+| Managed Identity Only (current preview baseline) | No per-user token; server uses MI for downstream calls | Simplifies backend, no OBO complexity | Loses per-user RBAC & auditing granularity; potential over-privileged MI |
```
localden, line 183: But how will that work in practice? For example, I connect your MCP server to my client. How does it know what identity to use for my use-case?

docs/Hosting.md
```diff
@@ -0,0 +1,417 @@
...
+| Brokered Auth (VS Code / WAM) | Leverage existing signed-in account in host environment | Smooth UX for first-party tools | Not portable to arbitrary MCP clients |
```
localden, line 184: This will only work for VS Code and VS :) (or other MSFT clients)

docs/Hosting.md
jongio: Den - Just so we know what we are saying no or yes to, could you let us know what it would mean to implement AS?

jongio (commented):
Reply to comment #9 (question about what it would mean to "implement AS").

Implementing our own Authorization Server (AS) instead of relying on Entra ID would entail a substantial scope with ongoing operational + security burden. High-level breakdown:

Core Protocol Surface
- OpenID Connect discovery endpoint (/.well-known/openid-configuration) exposing authorization, token, jwks_uri, issuer metadata.
- JWKS (JSON Web Key Set) hosting + automated signing key rotation (kid versioning, overlap / retirement policy, rollover announcement).
- Authorization Endpoint (Authorization Code + PKCE support, potentially Device Code, possibly future WebAuthn / passkey enhancements).
- Token Endpoint handling grant types: authorization_code, refresh_token, device_code (maybe), on-behalf-of / token exchange if we mimic OBO semantics internally.
- (Optional) Revocation (RFC 7009) and Introspection (RFC 7662) endpoints.

Client & Session Management
- Client registration (static config now; dynamic client registration adds more endpoints + validation logic).
- Policy & constraints per client (redirect URI allowlist, PKCE enforcement, public vs confidential, rate limiting, scope restrictions).
- User session management (auth cookies / state, consent records, CSRF protection for browser-based flows, anti-replay nonce tracking).

Token Lifecycle & Security
- Access & ID token issuance: signing (e.g., ES256/RS256), claims selection, audience scoping, issuer uniqueness, exp/nbf clock skew handling.
- Refresh token issuance & rotation (anti-reuse, sliding expiration strategy), revocation lists.
- Token exchange logic (if we simulate OBO) and internal caching keyed by (user, resource, scopes).
- Threat mitigations: PKCE mandatory for public clients, replay protection (nonce/hash binding), TLS enforcement, rate limiting, anomaly detection.

Data & Privacy
- Storage of user principals, consent, minimal logs (structured, PII scrubbing). Need retention & purge pipeline.
- NO long-term storage of raw tokens; prefer hashed references.

Operations & Compliance
- High-availability deployment (multi-instance + key distribution). Key rotation pipeline (automated generation, staged publish, old key retirement).
- Monitoring: auth success/failure metrics, latency, error codes, suspicious patterns (credential stuffing, token misuse).
- Security review / threat modeling (MITM, token replay, SSRF against metadata endpoints, JWT confusion attacks).
- Incident response runbooks (key compromise rotation, client credential revocation, abuse throttling).

Ecosystem Impact
- Every non-Microsoft MCP client would need to trust and integrate with our AS instead of the standardized Entra environment.
- We would be assuming liability for auth correctness; any vulnerability directly impacts downstream Azure access if we minted resource tokens.

Why We’re NOT Doing This Now (Preview Rationale)
1. Scope & Cost: This is a mini identity platform; large engineering + security investment and ongoing maintenance.
2. Risk Concentration: Increases attack surface; Entra already handles these concerns at scale.
3. Strategic Flexibility: Waiting preserves option to leverage future Entra features (dynamic client registration, token exchange enhancements) without migration burden.
4. Time-to-Preview: MI-only mode + future OBO leverages existing Entra semantics; faster path to customer value.
5. Governance: Internal reviews will likely prefer using established enterprise identity rather than bespoke issuance service.

Near-Term Actions (instead of custom AS)
- Define “principal context” object early (auth_mode, user_oid?, tenant_id, scopes) so logs & telemetry are ready for when OBO arrives.
- Add abstraction layer so inserting token validation & OBO exchange later does not alter tool implementations.
- Document MI-only limitations (coarse RBAC, aggregated auditing) with least-privilege guidance.
- Prepare risk/benefit comparison for pre-registered multi-tenant public client vs per-vendor registration.

Exit Criteria to Reconsider Building AS (if ever)
- Entra still lacks a required capability (e.g., dynamic registration) after agreed waiting period.
- Third-party clients require flows Entra cannot deliver (unlikely in near term).
- Security / compliance mandates custom claims pipeline not feasible via Entra.

Conclusion: We should explicitly state in the spec that building a bespoke Authorization Server is out-of-scope for Preview and likely for GA unless Entra platform gaps persist. Will add an Appendix note summarizing this reply so the decision trail is captured.

Let me know if you’d like additional depth (e.g., estimated engineering weeks by area, key rotation procedure details, or a draft threat model checklist).

docs/Hosting.md
joshfree: AS = Authorization Server that understands DCR; basically roll our own security sensitive code for the auth server; so that clients could dynamically register or unregister themselves (GET,PUT,DELETE /register), receive their {unique_client_id, access_token, mgmt_uris} to then begin using standard oauth with Entra ID

docs/Hosting.md
jongio: Okay, will remove this as an option.

docs/Hosting.md
```diff
@@ -0,0 +1,1181 @@
...
+* Attempting to invoke a non-safe tool while unauthenticated MUST yield a standardized error (e.g., `{"error":"auth_required"}`) without disclosing whether the tool exists outside safe mode.
+* Safe tools MUST tag responses with a context attribute (e.g., `"authMode":"none"`).
+
+### 20.7 Future Expansion Guardrails
+Any proposal to add more unauthenticated tools MUST provide: threat analysis, data classification, abuse scenarios, rate limit plan, rollback plan.
+
+### 20.8 Open Questions (Linked to Section 16)
+| Question | Link |
+|----------|------|
+| Should Foundry deployment listing remain excluded despite being non-mutative? | Section 16 (Demo Mode Scope) |
```
dem108, line 824: Yes, I think foundry deployment listing should remain exclude. It requires authentication to Azure and would take tenant and project endpoint from Foundry as parameters.
Instead, we plan to add a new tool GetModelDetail - like ListModels, this uses read-only public API that doesn't require authentication.

jongio:
Closing this in prep for new repo

